PR #61, opened 2015-11-09
Upgrade to Babel 6

d6u:
I tried to upgrade to Babel 6, but got stuck on two test cases. Need some help here. Thanks!


yyx990803:
Thanks for the PR - the first fail is actually because Babel 6 takes a long time to load and we didn't change mocha's default timeout. Also Babel 6 has a different module transform output, which needs some fixing... I'll merge this as-is and apply the changes necessary, thanks!

PR #81, opened 2015-11-28
Preserve line number for linters

rhyzx:
this PR will keep and comment all non-script source code in scripts output,
thus it will preserve line number and be friendly to linters(include file newLine ending).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -141,6 +141,7 @@ describe('vue-loader', function () {
       getFile('test.build.js.map', function (map) {
         var smc = new SourceMapConsumer(JSON.parse(map))
         getFile('test.build.js', function (code) {
+          console.log(code)
           var line
           code.split('\n').some(function (l, i) {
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -141,6 +141,7 @@ describe('vue-loader', function () {
       getFile('test.build.js.map', function (map) {
         var smc = new SourceMapConsumer(JSON.parse(map))
         getFile('test.build.js', function (code) {
+          console.log(code)
```
yyx990803: can we remove the log here please?

yyx990803:
Very cool! Just a small fix and LGTM.

yyx990803:
Actually, nvm I'll do that :) Thanks for the PR!

test/test.js
rhyzx: Oops, it's my mistake, thanks for your fixing.

PR #84, opened 2015-11-30
Resolved issue with CoffeeScript failing to build.

taylorzane:
This PR resolves an issue with CoffeeScript (and CS-like languages) due to the parser adding in `//` comments into the code passed to the respective loader, `coffee-loader` in this case.

Instead of passing commented code, I have modified the code to only pass the code between the `<script>` tags.

I have left the `commentScript` function in case it serves another purpose down the road.

@yyx990803 What was the original purpose for this function? It seems to cause unneeded behavior, since the ES6/Babel parser would ignore the commented lines anyway.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -74,9 +74,9 @@ module.exports = function (content) {
     var result
     if (type === 'script') {
       result =
-        commentScript(content.slice(0, start)) +
-        content.slice(start, end) +
-        commentScript(content.slice(end))
+        // commentScript(content.slice(0, start)) +
+        content.slice(start, end)// +
+        // commentScript(content.slice(end))
     } else {
       result = content.slice(start, end).trim()
     }
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -71,7 +71,15 @@ module.exports = function (content) {
     var start = node.childNodes[0].__location.start
     var end = node.childNodes[node.childNodes.length - 1].__location.end
 
-    var result = content.slice(start, end).trim()
+    var result
+    if (type === 'script') {
+      result =
+        commentScript(content.slice(0, start), lang) +
+        content.slice(start, end) +
+        commentScript(content.slice(end), lang)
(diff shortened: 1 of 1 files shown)


yyx990803:
Why would coffee-loader choke when there are comments? I'm not sure if I follow this.

The comments are added for correct source map line numbers: https://github.com/vuejs/vue-loader/pull/81

taylorzane:
Got it. The comments make sense then.

CoffeeScript uses the Ruby-style comments: `# this is a comment`.

CoffeeScript interprets a `//` as an empty RegExp.

taylorzane:
Which causes a parse error when coffee-loader encounters `</script>` because it sees that as an unclosed regular expression.

yyx990803:
Ahh I guess we will have to just use blank lines then...

taylorzane:
I had another idea that could preserve the code. I'll update my pull request accordingly.

taylorzane:
@yyx990803 I implemented a more robust solution. The `commentScript` function now detects which language is being used and uses the correct comment syntax.

lib/parser.js
```diff
@@ -91,14 +91,32 @@ module.exports = function (content) {
...
-function commentScript (content) {
+function commentScript (content, lang) {
   return content
     .split(/\n\r|\n|\r/g)
     .map(function (line) {
-      if (line.trim() === '') {
-        return line
-      } else {
-        return '// ' + line
+      switch (lang) {
```
yyx990803: I think we can check for `line.trim()` outside the switch, since it's the same case for all languages.

taylorzane: Very true. Seems extraneous.

yyx990803:
Very cool! I found out that there's a mis-configured eslint rule that makes the build fail, and just pushed a fix to `master`. Could you please submit a fresh PR based off latest master (also see the inline note above)? Thanks!

PR #85, opened 2015-11-30
Resolved issue with CoffeeScript failing to build

taylorzane:
This PR resolves an issue with CoffeeScript (and other non-standard languages) due to the parser adding in `//` comments into the code passed to the respective loader, coffee-loader in this case.

I have modified the `commentScript` function to use the correct commenting syntax for (most) all available webpack loaders.


yyx990803:
Thanks! :+1:

PR #98, opened 2015-12-09
Support "src" attribute for script tag

naumovs:
The intention of this PR is to have *.vue files as entry points for components, but the code separated across the js/html/css files and have all the benefits of vue-loader, such as hot-reload

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -153,10 +153,15 @@ module.exports = function (content) {
     })
 
     // add require for script
+    var script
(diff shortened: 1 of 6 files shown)


yyx990803:
Thanks, can you add an accompanying test case for this?

naumovs:
Added a simple test for an src import. Any other test cases needed?

yyx990803:
Looks good, thanks!

PR #111, opened 2015-12-29
add vue npm depend

iugo:


yyx990803:
Thanks! But `vue` should be a direct dependency instead of devDependency. So we need to install it separately with `--save`.

PR #205, opened 2016-04-13
Skip rewriter injection when loader has absolute path

### LINKED ISSUE #204: Module not found: Error: Cannot resolve 'file' or 'directory' ./../../../../../../../node_modules/extract-text-webpack-plugin/loader.js
See https://github.com/vuejs-templates/webpack/issues/41

oaleynik:
rewriterInjectRE matches single `html` and `css` words, so if loader is referenced via absolute path (extractTextPlugin case) and that path contains `html` or `css` words than rewriter will be injected right after them. This should not happen.

Fixes #204

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -96,9 +96,18 @@ module.exports = function (content) {
       // inject rewriter before css/html loader for
       // extractTextPlugin use cases
       if (rewriterInjectRE.test(loader)) {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks a LOT for finding the cause of the problem, I was in fact able to come up with a better regex though :)

Still, appreciate your effort in looking into this!

PR #208, opened 2016-04-19
preapre translation for multi languages

kazupon:


kazupon:
ping @vuejs/vue-loader-vueify

Issue #190, opened 2016-03-25
webpack CSS source map line numbers in the error messages


jpokrzyk:
CSS line numbers are incorrect in linters (stylint) and preprocessor error messages (webpack, stylus..)

Check out the line numbers. 
![image](https://cloud.githubusercontent.com/assets/4333958/14035730/e5b54c3c-f207-11e5-9b87-30f3764c8502.png)

Per Evan the source map line numbers need to be hacked. Or maybe somehow pad some blank lines.

I'm pretty busy over the next week. But if no one gets to it by then I'll probably take a stab at this.

Originally submitted here: https://github.com/vuejs-templates/webpack/issues/56

indirectlylit:
I'm noticing similar behavior. e.g. this .vue file:

```html
<template>
  test
</template>

<script></script>

<style lang="sass">
## invalid
</style>
```

generates this error:

```
    ERROR in ./~/css-loader!./~/vue-loader/lib/style-rewriter.js!./~/sass-loader?indentedSyntax!./~/vue-loader/lib/selector.js?type=style&index=0!./kolibri/plugins/learn/assets/src/learn-main.vue.html
    Module build failed: 
    <template>
    ^
          Invalid CSS after "": expected 1 selector or at-rule, was "## invalid {}"
          in /Users/d/Projects/learning_equality/repos/kolibri/kolibri/plugins/learn/assets/src/learn-main.vue.html (line 1, column 1)
     @ ./~/vue-style-loader!./~/css-loader!./~/vue-loader/lib/style-rewriter.js!./~/sass-loader?indentedSyntax!./~/vue-loader/lib/selector.js?type=style&index=0!./kolibri/plugins/learn/assets/src/learn-main.vue.html 4:14-303
```

Here it's even more confusing because it actually quotes text from the .vue file, taken from the wrong location, e.g.:

```
    <template>
    ^
```

indirectlylit:
@jpokrzyk - curious if that change helped with your issue?

Hucy:
): It does not seem to work
```
 "vue-html-loader": "^1.0.0",
    "vue-loader": "^8.5.2",
    "vue-style-loader": "^1.0.0",
```
and error:
```
ERROR in ./~/css-loader?sourceMap!./~/vue-loader/lib/style-rewriter.js!./~/stylu
s-loader!./~/vue-loader/lib/selector.js?type=style&index=0!./src/components/News
View.vue
[...]

PR #230, opened 2016-05-17
pad results with whitespace - re #190

indirectlylit:
Re: #190 – Padded the output to get line numbers accurate. Not sure if this is the correct strategy or not. Also did not test the effects on source maps.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -97,9 +97,7 @@ module.exports = function (content, filename, needMap) {
     var start = node.childNodes[0].__location.startOffset
     var end = node.childNodes[node.childNodes.length - 1].__location.endOffset
     var result
-    var lineOffset
     if (type === 'script') {
-      lineOffset = 0
       // preserve other parts as commenets so that linters
       // and babel can output correct line numbers in warnings
       result =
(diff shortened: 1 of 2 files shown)


indirectlylit:
Unfortunately, it looks like this change does not work because it doesn't take into account multiple `style` blocks. I'll look into that.

indirectlylit:
Actually that's not true: it does report proper line numbers, even with multiple style blocks. That said, the whole strategy feels a bit hacky, as though it should be handled with source maps or similar. Unfortunately that's outside my expertise.

Any feedback would be appreciated.

test/test.js
```diff
@@ -215,7 +215,8 @@ describe('vue-loader', function () {
     }), function (err) {
       expect(err).to.be.null
       getFile('test.output.css', function (data) {
-        expect(data).to.contain('h1 {\n  color: #f00;\n}\n\nh2 {\n  color: green;\n}')
+        console.log(data);
```
yyx990803: can we remove the log here?

lib/parser.js
```diff
@@ -128,7 +131,7 @@ module.exports = function (content, filename, needMap) {
           map.addMapping({
             source: filenameWithHash,
             original: {
-              line: index + 1 + lineOffset,
+              line: index + 1,
```
yyx990803: it seems after this change, `lineOffset` is only needed for the padding calculation, so it should be declared directly in the `else` block above.

indirectlylit:
Thanks for the comments - fixed.

This PR has improved parse and lint errors for us, but I'm not convinced it's the most correct solution to the problem. For example: when source maps are enabled, chrome dev tools shows the padded, style-only file rather than the original `.vue` file.

yyx990803:
Yeah, agreed. I'll merge this for now as a temporary workaround, and will see if we can get better sourcemap handling in the future. Thanks!

PR #240, opened 2016-05-24
Allow vue-loader to user PostCSS options as object and custom syntax

esiao:
I wanted to use SugarSS with PostCSS with vue-loader. It wasn't possible without modifying the style-rewriter and changing the way options works.

You still can use the 
`postcss: array of plugins`
`postcss: function that return array of plugins`

or the new one
```
postcss: {
  plugins: array of plugins,
  syntax: custom syntax, like sugarss
}
```

Allowing the use of sugarss or others syntax as custom postcss syntax with vue loader.

### DIFF
--- a/lib/style-rewriter.js
+++ b/lib/style-rewriter.js
@@ -40,8 +40,7 @@ module.exports = function (css, map) {
   var plugins = options.postcss
   if (typeof plugins === 'function') {
     plugins = plugins.call(this, this)
-  }
-  else if (typeof plugins === 'object') {
+  } else if (typeof plugins === 'object') {
     plugins = options.postcss.plugins
   }
   plugins = plugins ? plugins.slice() : [] // make sure to copy it
--- a/lib/style-rewriter.js
+++ b/lib/style-rewriter.js
@@ -40,7 +40,7 @@ module.exports = function (css, map) {
   var plugins = options.postcss
   if (typeof plugins === 'function') {
     plugins = plugins.call(this, this)
-  } else if (typeof plugins === 'object') {
+  } else if (typeof options.postcss === 'object' && options.postcss.plugins) {
     plugins = options.postcss.plugins
   }
   plugins = plugins ? plugins.slice() : [] // make sure to copy it
@@ -84,7 +84,7 @@ module.exports = function (css, map) {
   }
 
   // postcss options, for syntax
-  if (options.postcss.syntax) {
+  if (options.postcss && options.postcss.syntax) {
     opts.parser = options.postcss.syntax
   }
 
--- a/lib/style-rewriter.js
+++ b/lib/style-rewriter.js
@@ -41,6 +41,9 @@ module.exports = function (css, map) {
   if (typeof plugins === 'function') {
     plugins = plugins.call(this, this)
   }
+  else if (typeof plugins === 'object') {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR! I think it maybe a good idea to mix all options other than `plugins` into the `opts` object for processing. So you'd use

``` js
postcss: {
  plugins: [...],
  parser: sugarss
}
```

And it would also work in case any other plugins need different options.

esiao:
This is definitively the approach to go for, using the plugin object makes a lot of sense and it will be more modular at the end.

yyx990803:
@esiao yeah, that'd be cleaner. Could you update the PR to support that instead?

esiao:
@yyx990803 Yeah sure, should we keep the legacy version too? With options.postcss being an array or a function.

yyx990803:
@esiao yes, it seems perfectly feasible to do so and it's good to avoid breaking changes.

esiao:
@yyx990803 done. I'm checking if the user provided the options object for postcss and merging it with the one defined because `from` and `to`options should not come from the user.

yyx990803:
Great work! Thanks

esiao:
@yyx990803 I pushed an update, there was an error thrown when using sourcemaps but no postcss options, that strangely made it past the verification process.

yyx990803:
Yeah, I've fixed it in 8.5.2.

yuchuanxi:
@esiao how did you set  the `lang` attribute for `<style>`  in order to syntax highlight

esiao:
In the end I'm using it like this:

`<style lang="postcss">`

And in webpack config

```
  vue: {
    loaders: {
      postcss: TARGET === 'build' ? ExtractTextPlugin.extract('style-loader', 'css-loader!postcss-loader?parser=sugarss') : 'style-loader!css-loader!postcss-loader?parser=sugarss'
    },
    autoprefixer: false
  },
```

But the syntax highlighting is broken since I'm using SugarSS syntax on what the Atom vue plugin considers as postcss syntax. The solution would be to use a custom lang like `lang="sugarss"` which means changing the loaders to `sugarss` and implement support in the Atom vue plugin, but I don't have the time for that for now.

yuchuanxi:
@esiao thanks, my situation like you ...

wprater:
@esiao do I have to override the loader for postcss as you've shown when using `<style lang="postcss">`?  I've tried to use the `lang` for  postcss, but get an error `... is not a PostCSS plugin`

esiao:
@wprater no, what I've shown here is the final solution I used on my project, I also had some external postcss files to process.

If you want to use postcss with vue-loader all you need to do is pass postcss options to vue config as explained here :
http://vue-loader.vuejs.org/en/features/postcss.html

wprater:
@esiao I have no issues using postcss, just the `lang` attribute causes errors.  probably a bug, so I'll file it elsewhere.

octref:
Hello, for people in this thread:

Is the current state people have to use `<style lang="postcss">` for both postcss and SugarSS?

For my VSCode extension, I have postcss syntax highlight for Vue SFC. I'd like to have SugarSS too, but if there is no way to tell postcss away from SugarSS it's impossible to write a grammar for it.

kodekrendel:
Whould love to see full SugarSS support as well!

PR #241, opened 2016-05-24
unit tests in single file (fixes #223)

paulpflug:
adds the possibility to add unit tests to the vue file - to keep together what belongs together

This is rather large. A short overview of the content:

- removed peer deps to remove errors when not using babel (see https://github.com/vuejs/vueify/issues/81)

- includes improved changes from #230 + missing unit test for style sourcemap

- bumped all deps

- moved source map test to use coffee-loader instead of babel-loader (which won't work as expected and failing the test)

- moved mocha tests to karma where possible to show of the new feature

missing: documentation - I'm willing to provide these if this gets accepted.

example for testing a unmounted component:
```vue
<script>
module.exports = {
  data () {
    return {
      msg: 'Hello from Component A!'
    }
  }
}
</script>
<template>
  <h2 class="red">{{msg}}</h2>
</template>

<style>
comp-a h2 {
  color: #f00;
}
</style>

<script test>
var module = window.vueTestComp
var styles = window.document.querySelectorAll('style')
var style = styles[styles.length-1].textContent
describe("vue-loader", function () {
  it('basic', function () {
    expect(module.template).to.contain('<h2 class="red">{{msg}}</h2>')
    expect(module.data().msg).to.contain('Hello from Component A!')
    expect(style).to.contain('comp-a h2 {\n  color: #f00;\n}')
  })
})
</script>
```

example for testing a mounted component:
```vue
<template>
  <h2 class="red">{{msg}}</h2>
</template>

<script>
module.exports = {
  data () {
    return {
      msg: 'Hello from Component A!'
    }
  }
}
</script>

<script test="karma">
var module = window.vueTestComp
describe("vue-loader", function () {
  it('karma', function () {
    expect(module.$el).to.exist
    expect(module.msg).to.contain('Hello from Component A!')
  })
  after(function () {
    module.$destroy(true)
  })
})
</script>
```

PR #244, opened 2016-05-26
removed peer dependencies

paulpflug:
to allow error free npm install when not using style / hot-reload or babel
#241

https://github.com/vuejs/vueify/issues/81


kazupon:
Close (outdate)

PR #245, opened 2016-05-26
test and improvement for #230

paulpflug:
replaced commenting other parts of vue file with newlines in sourcemap + test for stylus sourcemap
#241


kazupon:
Close (outdate)

PR #256, opened 2016-06-16
shold emit webpack Error rather than throwing

Eden-Sun:
prevent webpack crashing when templateLoader compiled errors.
https://github.com/vuejs/vue-cli/issues/111

### DIFF
--- a/lib/template-loader.js
+++ b/lib/template-loader.js
@@ -5,6 +5,7 @@ var extname = require('path').extname
 module.exports = function (content) {
   this.cacheable && this.cacheable()
   var callback = this.async()
+  var emitError = this.emitError
   var opt = loaderUtils.parseQuery(this.query)
   var vue = this.options.vue
   if (vue && vue.template) {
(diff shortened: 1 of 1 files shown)


lib/template-loader.js
```diff
@@ -27,18 +28,18 @@ module.exports = function (content) {
   }
 
   if (!cons[opt.engine]) {
-    throw new Error(
+    emitError(new Error(
```
yyx990803: We should just call `callback` here with the error as the first argument. Also need to return early, so `return callback(new Error('...'))`

lib/template-loader.js
```diff
@@ -27,18 +28,18 @@ module.exports = function (content) {
...
+    ))
   }
 
   // for relative includes
   opt.filename = this.resourcePath
 
   cons[opt.engine].render(content, opt, function (err, html) {
     if (err) {
-      throw err
+      emitError(err)
```
yyx990803: Same as above

lib/template-loader.js
```diff
@@ -27,19 +27,20 @@ module.exports = function (content) {
   }
 
   if (!cons[opt.engine]) {
-    throw new Error(
+    callback(new Error(
```
yyx990803: we need to return here otherwise the code below will throw another error.

yyx990803:
Thanks!

PR #267, opened 2016-06-23
Add support for options.query of babel-loader

zhulongzheng:
It seems like that if you add `!!` before `babel-loader`, the options(including query) of `babel-loader` will be ignored by webpack, but the `babel:` option and `.babelrc` still works.
This PR just makes the `options.query` of `babel-loader` work.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -46,10 +46,20 @@ module.exports = function (content) {
     process.env.NODE_ENV !== 'production' &&
     options.cssSourceMap !== false
 
+  var babelQuery = ''
+  if (hasBabel) {
+    var globalLoaders = this.options.module.loaders
+    for (var i = globalLoaders.length - 1; i >= 0; i--) {
+      if (globalLoaders[i].loader === 'babel') {
+        babelQuery = globalLoaders[i].query ? '?' + JSON.stringify(globalLoaders[i].query) : ''
+      }
+    }
+  }
+
   var defaultLoaders = {
     html: templateCompilerPath + '?id=' + moduleId,
     css: styleLoaderPath + '!css-loader' + (needCssSourceMap ? '?sourceMap' : ''),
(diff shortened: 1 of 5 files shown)


lib/loader.js
```diff
@@ -46,10 +46,20 @@ module.exports = function (content) {
     process.env.NODE_ENV !== 'production' &&
     options.cssSourceMap !== false
 
+  var babelQuery = ''
+  if (hasBabel) {
+    var globalLoaders = this.options.module.loaders
+    for (var i = globalLoaders.length - 1; i >= 0; i--) {
+      if (globalLoaders[i].loader === 'babel') {
```
fnlctrl: `babel-loader` is also valid, should also be supported.

fnlctrl:
Actually, after reading the tests provided in this PR, I've found that setting `babel` on webpack.config alone does the trick, without any need to modify vue-loader.

Interestingly, babel-loader reads `this.options.babel` as globalConfig (see [this line](https://github.com/babel/babel-loader/blob/master/index.js#L79)), but it was never documented, and I haven't found any actual usage of it after searching on google.
 
So I suppose we can close this, but add a note to the docs that if you want to configure babel loader, just set `babel` option on webpack config.

test/test.js
```diff
@@ -79,6 +79,63 @@ describe('vue-loader', function () {
...
+            loader: loaderPath
+          },
+          {
+            test: /\.js?/,
+            loader: 'babel',
+            exclude: /node_modules/
+          }
+        ]
+      },
+      babel: {
```
fnlctrl: Setting `babel` alone will do the trick, without the need to modify vue-loader.

yyx990803:
Closing since a dedicated `.babelrc` is the recommended approach, and magically sniffing the query of another loader rule just doesn't sound right. Thanks for the PR though and sorry for letting it hang for so long.

PR #268, opened 2016-06-25
Handle partial tag (fix #134)

### LINKED ISSUE #134: Suggestion: partial definition tag
Would be nice to able to define component partials in a similar way we're defining the `template`, with a dedicated `<partial>` tag.

Feels wrong to have `template` code like this, defined in the `script` section, in the context of a `.vue` file:

partial-tag

Instead of the above, we could do:

partial-tag-suggestion-v2

Which in my opinion, feels much better.

Or, since a partial is essentially a template, without introducing a new `partial` tag which might be confusing, maybe just add support for multiple `template` tags, and a `partial` property, something like:

##### Main template

```html
<template>
   <partial name="dropdown-area-content"></partial>
</template>
```

##### Partial template

```html
<template partial="dropdown-area-content">
    ...
</template>
```

noirbizarre:
This PR allow to define partials with a `<partial>` tag as propposed in #134 allowing to write:
```xml
<template>
  <div>
    <partial name="my-partial"></partial>
  </div>
</template>

<partial name="my-partial">
  <div>Hello</div>
</partial>
```

The partials are injected into the `partials` option under the `name` key (ie. `my-partial` in the previous example)

Follow the same rule as `<template>` tag:
- template syntax specified with `lang` attribute
- can be loaded from another file with  `src` attribute

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -7,6 +7,7 @@ var path = require('path')
 
 var defaultLang = {
   template: 'html',
+  partial: 'html',
   style: 'css',
   script: 'js'
 }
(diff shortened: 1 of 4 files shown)


yyx990803:
Thanks for the PR - but I don't think `vue-loader` should support it. See https://github.com/vuejs/vue-loader/issues/134#issuecomment-229184502

PR #290, opened 2016-07-25
Include upstream loaders in require statements

NeonPaul:
Generated require statements should include any loaders that fall between vue-loader and the source file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ There are also some example projects:
 
 - [vue-loader-example](https://github.com/vuejs/vue-loader-example/)
 - [vue-hackernews](https://github.com/vuejs/vue-hackernews)
+
+## License
+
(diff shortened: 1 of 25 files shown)


yyx990803:
What's the use case for this? I don't think this is necessary because the require generation functions are used to extract parts from the raw `*.vue` file, any remaining requests are applied to the final output instead of individual parts.

NeonPaul:
We have components packaged as individual npm modules, each accompanied by a README.md with code examples in vue file format. I'm writing a loader to extract these code examples, parse them as vue code, and run them in order to (a) make sure they're up to date and correct and (b) avoid having to write the examples once in in the readme and once more in our demonstration project.

kazupon:
Close (outdate)

PR #308, opened 2016-08-09
Custom Rewriters

Xanderite:
Adds option to disable style rewriters, or use custom ones.

Example use case:
- building uncompiled preprocessor code.
  (we want to build skinnable widgets, so outputting a stylus file for a widget is a nicer way than rebuilding the whole project (many widgets) on each style variable change)
- implement scoping rules differently
  (in the above scenario keeping scoped styles requires this functionality)


kazupon:
Close (outdate)

PR #312, opened 2016-08-13
One change - Line 5

aarmand:
'with' is changed to 'which'


yyx990803:
Thanks but this PR seems out of date now after a branch move. Feel free to re-submit against 1.0 branch.

PR #316, opened 2016-08-17
Add ability to set `transformToRequire` option under `vue` in webpack…

fnlctrl:
Add ability to set `transformToRequire` option under `vue` in webpack config,
This provides the missing feature from [html-loader](https://github.com/webpack/html-loader) to specify which tag-attribute combination to be transformed to `require()`.

`transformToRequire` Defaults to {img: 'src'}.

Example usage: (in webpack.config.js)
```javascript
vue: {
  transformToRequire: {
    foo: 'bar'
  }
}
```
will transform `<foo bar="..."/>` to `<foo bar="require(...)"/>`

### DIFF
--- a/lib/template-compiler.js
+++ b/lib/template-compiler.js
@@ -8,7 +8,7 @@ var hotReloadAPIPath = normalize.dep('vue-hot-reload-api')
 var defaultTransformToRequire = {
   img: 'src'
 }
-var transformToRequire = Object.assign({}, defaultTransformToRequire);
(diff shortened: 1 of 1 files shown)


yyx990803:
Maybe also a good idea to make it support an array of attrs:

``` js
vue: {
  transformToRequire: {
    foo: ['bar', 'baz']
  }
}
```

fnlctrl:
Yeah, implementing it now.

fnlctrl:
Done.

PR #322, opened 2016-08-23
Provide default for name option

chrisvfritz:
Use the filename to set a component's name option, if one isn't provided already. This allows more helpful debugging messages by default.


chrisvfritz:
@yyx990803 Not sure whether I really need to check whether `__vue_options__` is an object here. Can that be considered guaranteed at this point?

yyx990803:
It should be an object at that point, so we can remove the object check. Also can we add a test case somehow?

chrisvfritz:
@yyx990803 Object check removed and tests added.

lib/loader.js
```diff
@@ -180,6 +180,10 @@ module.exports = function (content) {
     // constructor export interop
     'if (typeof __vue_options__ === "function") {\n' +
     '  __vue_options__ = __vue_options__.options\n' +
+    '}\n' +
+    // add name based on filename if not already set
+    'if (!__vue_options__.name) {\n' +
```
yyx990803: This `if` check seems redundant

PR #323, opened 2016-08-25
Add scoped attribute let component css-rules to overwrite the child c…

hinablue:
When I need to overwrite 3rd-party components style, if the parent is "scoped", all the child component's styles (if they are not scoped) cannot be overwrite cause style rules,

```
<template>
  <div class="parent-component">
    <child-component></child-component>
  </div>
</template>
<style scoped>
.parent-component .child-component {
  max-width: 100%;
}
</style>
``` 

```
<template>
  <div class="child-component">
  </div>
</template>
<style>
.child-component {
  max-width: 50%;
}
</style>
``` 

The parent component's style will be rendered to,

```
.parent-component .child-component[_v-d5346ea8] {
  max-width: 100%;
}
```

And child component's style will be rendered to,

```
.child-component {
  max-width: 50%;
}
```

Cause the css selector rule, the `.parent-component .child-component[_v-d5346ea8]` cannot apply to the child component (The rule does not match).

So I modified the style-rewriter to check if `[scoped]` attribute exists in parent component, do not apply the scope id after the selector like this,

```
<style scoped>
.parent-component[scoped] .child-component {
  max-width: 100%;
}
```

Rendered to,

```
.parent-component[_v-d5346ea8] .child-component {
  max-width: 100%;
}
```

So I can change the scope id position that can overwrite the child component's styles.

Also, It ONLY CAN BE OVERWRITE THE FIRST LEVEL OF CHILD COMPONENTS cause it cannot got the second level component's scope id in style-rewrite.

### DIFF
--- a/docs/en/README.md
+++ b/docs/en/README.md
@@ -31,7 +31,7 @@ But Webpack can do more than that. With "loaders", we can teach Webpack to trans
 - Transpile ES2015, CoffeeScript or TypeScript modules into plain ES5 CommonJS modules;
 - Optionally you can pipe the source code through a linter before doing the compilation;
 - Transpile Jade templates into plain HTML and inline it as a JavaScript string;
-- Transpile SASS files into plain CSS, then convert it into a JavaScript snippet that insert the resulting CSS as a `<style>` tag;
+- Transpile SASS files into plain CSS, then convert it into a JavaScript snippet that inserts the resulting CSS as a `<style>` tag;
 - Process an image file referenced in HTML or CSS, moved it to the desired destination based on the path configurations, and naming it using its md5 hash.
 
 Webpack is so powerful that when you understand how it works, it can dramatically improve your front-end workflow. Its primary drawback is verbose and complex configuration; but with this guide you should be able to find solutions for most common issues when using Webpack with Vue.js and `vue-loader`.
--- a/docs/en/features/postcss.md
+++ b/docs/en/features/postcss.md
@@ -15,6 +15,7 @@ module.exports = {
 }
 ```
 
+<<<<<<< HEAD
 For Webpack 2.x:
 
 ``` js
--- a/docs/en/workflow/linting.md
+++ b/docs/en/workflow/linting.md
@@ -2,7 +2,7 @@
 
 You may have been wondering how do you lint your code inside `*.vue` files, since they are not JavaScript. We will assume you are using [ESLint](http://eslint.org/) (if you are not, you should!).
 
-You will also need the [eslint-html-plugin](https://github.com/BenoitZugmeyer/eslint-plugin-html) with supports extracting and linting the JavaScript inside `*.vue` files.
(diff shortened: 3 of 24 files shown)


catalsdevelop:
good, i have also encounter those problems. shadow dom has '::shadow' or '/deep/' to set child component style.

yyx990803:
Hmm, the root of the child component should be style-able from the parent, because it gets both the scope attribute from the parent AND the child. If it is not, then it seems to be a bug.

jvmccarthy:
Thanks for the response @yyx990803 and for your great work with vue!

Let me try to clarify my use case. I'm including an svg icon dynamically in a scoped component like so,
```html
<div class='icon'>
  {{{ getIcon() }}}
</div>
```
where `getIcon()` returns the svg (loaded with the html loader). 

The generated markup looks like this,
```html
<div class='icon' _v-1fd1522a>
  <svg> <!-- (with additional attributes and child path elements) -->
  </svg>
</div>
```

I'm trying to style the fill color of the svg like this,
```css
.icon svg {
  fill: #000000;
}
```
which is getting generated like this,
```css
.icon svg[ _v-1fd1522a] {
  fill: #000000;
}
```

As a result, the svg is not styled as desired. I've got a child element that is not scoped inside a parent element that is scoped, and I want to style the child under the parent. 

Can you clarify how the child component should be styled in a such a case?

Thanks again for efforts and for sharing vue with us all!

hinablue:
@yyx990803 

The style-rewriter in loader always append the attribute `_v-xxxxxxx` in the end of the selector, so the parent cannot style-able the child component.

Just like @jvmccarthy use case, the parent component will be generated the style like,

```
.icon svg[_v-xxxxxx] {
    fill: #000000;
}
```

And the child component will never fit the css selector rule like above. It should be,

```
.icon[_v-xxxxxx] svg {
    fill: #000000;
}
```

I also use some 3rd party component like vue-strap, If I wrote this,

```
<div class="parent-component">
    <v-alert></v-alert>
</div>
```

And HTML will be generated to,

```
<div class="parent-component">
    <div class="alert"> ... </div>
</div>
```

And in parent component style like this,

```
<style scoped>
.parent-component .alert {
}
</style>
```

It will be generated to,

```
.parent-component .alert[_v-af39gh] {
}
```

And this style rule will not fit the DOM Tree.

yyx990803:
I think it breaks the consistency when `<style scoped>` can in fact affect components out of the current scope, and the usage can be a bit obscure to understand.

In fact you can just mix a non-scoped `<style>` if you want normal cascading selectors - of course now you need to make sure there is a globally unique class so it doesn't affect others, but I think this is not really something `scoped` should be dealing with.

jvmccarthy:
Thanks again for considering this. I was able to workaround my specific issue by using the [Cascading SVG Fill Color](https://css-tricks.com/cascading-svg-fill-color/) trick from css-tricks.

PR #345, opened 2016-09-10
template opt prop - fixes #342

### LINKED ISSUE #342: resolve `require` in <template>
This is possible with [pug-loader](https://github.com/pugjs/pug-loader) but not with `vue-loader`

``` pug
div
  img(src=require("./my/image.png"))
```

This also would be very handy:

``` pug
- sitemap = require("./sitemap.json")
```

It would allow to use the templating engine iteration mechanism instead of `v-for` wherever `v-for` isn't suited.

paulpflug:
allows following syntax:

``` vue
<template lang='pug' opt='{"pretty":true}'>
</template>
// and
<template lang='pug' opt='options.json'>
</template>
```

the resolved opt object will be forwarded to the templating engine.

With unit test but without documentation (yet)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ There are also some example projects:
 
 - [vue-loader-example](https://github.com/vuejs/vue-loader-example/)
 - [vue-hackernews](https://github.com/vuejs/vue-hackernews)
+
+## License
+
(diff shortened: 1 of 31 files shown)


kazupon:
Close (outdate)

PR #395, opened 2016-10-11
Update production.md (#373)

jbruni:


jbruni:
"Deploying for Production" section does not exist anymore at
http://vuejs.org/guide/single-file-components.html#Deploying-for-Production

It has been moved to its own page at
https://vuejs.org/guide/deployment.html

PR #396, opened 2016-10-11
Provide default for name option

chrisvfritz:


chrisvfritz:
@yyx990803 This should be ready now.

PR #421, opened 2016-10-19
remove css-loader from peerDependencies

Phieo:
Since npm3+,it will not install peerDependencies automatically.when I run `npm shrinkWrap --dev`,it shows the error:**peer invalid: css-loader@*, required by vue-loader@9.7.0**


yyx990803:
You should manually install the peerDep with `--save` before running shrinkwrap.

PR #429, opened 2016-10-27
Update advanced.md - match style of extract-css.md

jbruni:


jbruni:
Just compare the Webpack config examples headers at:
- http://vue-loader.vuejs.org/en/configurations/advanced.html
- http://vue-loader.vuejs.org/en/configurations/extract-css.html

The bold heading of "extract-css" doc sections is better for reading than the "advanced" doc, in my opinion.

In any case, the styles between these pages should match.

PR #422, opened 2016-10-19
[WIP]CSS modules

YiSiWang:
@yyx990803 
2.x uses `vue-template-compiler`, which doesn't expose non standard props (`module`) on `<style>` tag.

Should we use `data-module` instead or modify `vue-template-compiler` to expose non standard props?

I prefer the later cause `scoped` was removed from standard.

PR #442, opened 2016-11-03
Support css modules at front/server side

PowerfulPony:
work depends on the following revisions

* Add support attribute module - vue-template-compiler
https://github.com/vuejs/vue/pull/4091
* Do not add a handler on the server
https://github.com/vuejs/vue-style-loader/pull/1

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -181,7 +181,7 @@ module.exports = function (content) {
     parts.styles.forEach(function (style, i) {
       var module = style.module
 
-      if(!isServer || module) {
+      if (!isServer || module) {
         var requireString = style.src
           ? getRequireForImport('styles', style, style.scoped)
           : getRequire('styles', style, i, style.scoped)
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -77,7 +77,7 @@ module.exports = function (content) {
       // disable all configuration loaders
       '!!' +
       // get loader string for pre-processors
-      getLoaderString(type, part, scoped) +
+      getLoaderString(type, part, index, scoped) +
       // select the corresponding part from the vue file
       getSelectorString(type, index || 0) +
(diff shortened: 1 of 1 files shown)


PowerfulPony:
Just due to the need to use "css modules" at SSR

These changes are possible because of @YiSiWang
https://github.com/vuejs/vue-loader/pull/422
https://github.com/vuejs/vue-loader/pull/331

yyx990803:
Is there any change in this PR that is not covered / needs to be addressed in #422 ? - would it be better if we merge the two dependent PRs and then merge #422 instead?

PowerfulPony:
+It works with SSR
+The server-bundle fall only css modules
+css modules are named according to the attribute of the module
Because styles are loaded through vue-style-loader, it is necessary to fix the fragment associated with the addition of hot.reload script https://github.com/vuejs/vue-style-loader/pull/1
Cause the server to will throw an error because the browser checks

In this PR are implemented that was available in version 8, there is also support for SSR.
Important to me only anyside support css-module, since using SSR.

It is also possible to fully restore the documentation of version 8
https://github.com/vuejs/vue-loader/pull/331

I am for the result!
I just want to pay attention to the css-modules, in the name of the module, and ssr

YiSiWang:
@PowerfulPony Thank you for pointing out problems with SSR. But I think it's better to drop `style-loader` when in SSR.

1. Faster && less code
2. No need for vuejs/vue-style-loader#1

I've done it in #422 with test case.

PowerfulPony:
@YiSiWang 
Yes, such a solution is better
https://github.com/vuejs/vue/pull/4091 Please pay attention, it is already possible to use the attribute in dev branch

yyx990803:
In that case, closing this in favor of #422

PR #446, opened 2016-11-06
use `Object.assign` to merge cssModules configuration `query` with `option`

JounQin:
So that more options can be used instead of listing them specifically like `camelCase`.

And maybe there is no need to override the undefined localIdentName option which is passed customly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ npm-debug.log
 test/output
 docs/_book
 .DS_Store
+.idea
+*.iml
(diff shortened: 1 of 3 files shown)


yyx990803:
What's an example use case for this?

JounQin:
If there is a class named `a-b`, and there is no `camelCase` option or it is false, we can only get the generated class name by `$style['ab-cd']`.

But when the `camelCase` is set to be true, we can get the class name by `$style.abCd` which is much better to write and 'js-style'.

yyx990803:
Thanks!

Issue #464, opened 2016-11-14
vue-loader breaks since Webpack 2.1.0-beta.26


riovir:
https://github.com/webpack/webpack/releases/tag/v2.1.0-beta.26
The breaking change now forces -loader to be used. Since vue-loader internally also uses loaders the problem might be caused by this change. See:

Error: Module '....../node_modules/less/index.js' is not a loader

riovir:
In the loader.js the line 80 the following horrible hack solves the problem:
getLoaderString(type, part, index, scoped).replace(/!less!/, '!less-loader!') 

Naturally this is *not* a proposed fix, merely a step to pin down the issue.

riovir:
An actual proposal here for a fix:
@line 155
return loader + '!' + rewriter + lang + '-loader!'

thorning:
I had this problem using `lang="stylus"` and a (temporary) fix for me was to change it to `lang="stylus-loader"`, I also had to update loader references in the webpack config

ylhuang0423:
I' m using Webpack 1.13.3 (npm version)
`resolveLoader.alias` is set in webpack.config.js (i.e `sugarss: 'postcss-loader?parser=sugarss'`)
The fix would break my setting, because `sugarss` becomes `sugarss-loader` implicitly.

Webpack 2.1.0-beta.26 (Webpack) is still a beta version (even Webpack 2.0 is beta).
The fix may effects lots of people that still using Webpack1.0 with Vue.

riovir:
@thorning, I tested the "fix" with both Jade and Pug. For templates "raw" changes to "raw-loader" the engine query param stays the same. See:
`case 'template':
          return defaultLoaders.html + '!' + templateLoaderPath + '?raw-loader&engine=' + lang + '!'`

@ylhuang0423, Breaking `resolveLoader.alias` is a different matter though.

riovir:
At the moment I don't have good proposal on how to recognize when there would be an alias collision, other than adding a query / option parameter explicitly turning the -loader fix on or off. I wonder how a loader can tell when it's necessary to suffix the lang in a safe way...

ylhuang0423:
@riovir, Maybe try...catch to require the loader (like what hasBabel do) could be a possible solution. 
If a loader is existed indeed then suffix -loader to it, or it'll be a [Inline Loader Request](https://vue-loader.vuejs.org/en/configurations/pre-processors.html) or alias.

But it's not a perfect solution for some special cases.
[...]

PR #465, opened 2016-11-14
[Fix for Webpack 2 Beta 26+] Explicitly declare inferred loaders by their full name (#464)

riovir:
* In Webpack 2 loaders have to be explicitly declared, the "-loader" part of their name is no longer implied.

This change assumes the use case where lang already ended with '-loader' is not typical or preferred.


yyx990803:
Thanks!

PR #468, opened 2016-11-15
raw is not loader

### LINKED ISSUE #466: Templates written in pug result in one root element warning
After upgrading to 9.9.1, I started getting errors one pug templates.  The templates had only one root element but failed compiling with the error that component templates should have only one root element.

cxr29:
fix #466


yyx990803:
oops, sorry, just noticed this... already fixed in 9.9.3

PR #474, opened 2016-11-18
update `vue-template-compiler`

cnwhy:


yyx990803:
This is not necessary because it uses a semver range. https://github.com/npm/node-semver#caret-ranges-123-025-004

PR #495, opened 2016-11-25
When name isn’t specified use the filename as the component name.

bfanger:
Improves the Vue.js devtools experience, no more “< Anonymous Component >”


blake-newman:
Makes sense. Can't see any issues with this.

Issue #457, opened 2016-11-10
[feature request][discussion welcome] Component Translation Section


gbezyuk:
Reusable components could be used in different language environments. That being said, we could benefit from having modular translations, by adding another section in single-file components we already use, probably JSON-based:

<translation>

```json
{
  "Select a color": {
    "ru": "Выберите цвет",
    "de": "wählen Sie eine Farbe"
  }
}
```
</translation>

kazupon:
Close (implementing custom blocks with #499)

Issue #423, opened 2016-10-20
Vue-Loader for generating styleguides


iamkether:
I've been thinking about styleguides over the last few days after a discussion in the Vue slack channel with Alexander Hofstede. I'm not entirely sure if this would actually be something that would fit inside of the scope of Vue-Loader, but I think this is where it makes _the most sense_ to at least start the discussion because the syntax I'm thinking of is quite related to Vue-Loader's existing component spec. I published a [gist](https://gist.github.com/iamkether/7f03c7f4397dee6a92e646e6aa718e05) outlining the deeper ideas and reasoning, but the general idea is to add one more tag to the Vue-Loader component spec:

```
<styleguide lang="markdown">
  This area would allow virtually any kind of markdown text, comments, etc. 
</styleguide>

<template>
  // ...
</template>

<script>
  // ...
</script>

<style>
  // ...
</style>
```

And then we could use the markup from that tag to generate a component tree/styleguide based on the structure of existing projects.

aphofstede:
Interesting. I think keeping the documentation close to the component it describes is an excellent idea (for Vue); It seems to fit the general paradigm of small scopes and self-contained components. A similar argument could be made for including tests in components' Vue files as well. 

I guess the primary discussion would come down to "how much crap do we want people to put in their Vue components' source?". I don't know the history and vision of Vue well, so I'd be interested to hear their point of view.

CSS- / JS doc can definitely be parsed to enrich the generated style guide; same goes for the Vue component's props.

On a side note / FYI: In [semantic-react](https://github.com/hallister/semantic-react), which uses [react-styleguidist](https://github.com/sapegin/react-styleguidist), they put their components in [subfolders](https://github.com/hallister/semantic-react/tree/master/src/components/elements/divider) and have their tests and examples (style guide) next to, but not in, their components.

iamkether:
@aphofstede quite valid points and after looking at the level of documentation for some of the components in that semantic-react library, I can tell I was underestimating the level of complexity for the documentation requirements.

egoist:
This would play nice with https://github.com/vue-play/vue-play 😃 It accepts an `readme` property to display something like `styleguide`

iamkether:
[...]

Issue #413, opened 2016-10-15
Support theming through attributes on style tags


zephraph:
``` html
<style lang="stylus">
  p
    color black
</style>

<style lang="stylus" theme="light">
  p
    color white
</style>
```

I'd like to be able to create component themes given an attribute name. Idealistically no theme attribute would roll up into the default css file and style blocks with the theme attribute would roll up in a css file of that name. 

I'm willing to do the leg work on this and put in a PR if it sounds reasonable.

nblackburn:
I don't feel like this is relevant to vue-loader or something universally useful.

zephraph:
You might be right in that it deals with more than vue-loader. The `theme` attribute, in particular, is unimportant. What I'd like to be able to do is have attributes pass through from vue-loader to the style blocks that eventually get injected into the browser. 

Generating separate stylesheets entirely would likely be handled by a modified version of the extract text plugin or something.

kazupon:
Close (implementing custom blocks with #499)

Issue #223, opened 2016-05-11
including unit tests in vue component


paulpflug:
For unit testing of standalone components, I have a `vue` environment file where I load the component and interact with it. I often find myself switching between the environment file and the file containing the unit test fore and back..

This would very much simplify the unit testing.
this: 

``` vue
<template><someComp></someComp></template>
<script>
module.exports = {
  components: {
    someComp: // the component to test
  }
  // other stuff to interact with someComp
}
</script>
<script test="karma">
describe("comp", function() {
  it("should work", function() {
    //the test
  })
})
</script>
```

compiles to this:

``` js
module.exports = {
  components: {
    someComp: // the component to test
  }
  // other stuff to interact with someComp
};

if (module.exports.__esModule) module.exports = module.exports.default
;(typeof module.exports === "function"? module.exports.options: module.exports).template = "<someComp></someComp>"

// stripped in production build
if (window.__karma__ !== "null") {
  karmaTestComp = Vue.extend(module.exports)
  karmaTestComp = new karmaTestComp()
  document.body.appendChild(karmaTestComp.$mount().$el)
  describe("comp", function() {
    it("should work", function() {
      //the test
    })
  })
  document.body.removeChild(karmaTestComp.$el)
  karmaTestComp.$destroy()
}
```

Karma would only need to provide `Vue` and webpack-load the environment vue file

kazupon:
Close (implementing custom blocks with #499)

PR #499, opened 2016-11-25
Add support for custom blocks in .vue files

asselin:
This builds on https://github.com/vuejs/vue/pull/4157 adding support in vue-loader for custom blocks.  If a custom block is found in a .vue file, the loader for it will be looked up in the webpack vue configuration block just like a language is for script/style/template.

I think this improvement could probably cover the requests in some other feature requests, such as #457, #423, #413, #223, and having the ability in a UI kit project to specify (e.g.) Sass in a .vue file for a GUI component that is meant to be extracted and concatenated, and compiled into CSS by the application that pulls in and uses the UI kit.

This is a pretty simplistic implementation-- if there are issues that you think should be addressed in order to merge this, please let me know.


lib/loader.js
```diff
@@ -271,6 +273,20 @@ module.exports = function (content) {
       '__vue_options__.staticRenderFns = __vue_template__.staticRenderFns\n'
   }
 
+  // add requires for customBlocks
+  if (parts.customBlocks && parts.customBlocks.length) {
```
yyx990803: We should avoid adding any code to the file unless there is a matching loader for the custom block.

lib/loader.js
```diff
@@ -155,6 +155,8 @@ module.exports = function (content) {
           return loader + '!' + rewriter + ensureBang(ensureLoader(lang))
         case 'script':
           return injectString + ensureBang(ensureLoader(lang))
+        default:
+          return ensureBang(loaders[type])
```
yyx990803: This would probably throw if `loaders[type]` doesn't exist

yyx990803:
Thanks for the PR! A few issues that need to be fixed, and ideally we need a test case for this too.

asselin:
Thanks for the feedback!  I fixed the 2 issues and added them to the PR.  I'll work on adding some tests and updating the documentation one night this week.

asselin:
Hmmm.  I think I messed up my rebase... I'll try fixing that so the patch is cleaner...

asselin:
FYI, I haven't forgotten about building unit tests for this PR; I've just been slammed with a number of things demanding my time right now.  I still intend to finish this up, targeting the week of Dec 26-30.

asselin:
Unit tests added, please let me know if there are any other changes required.

iamkether:
@asselin this is looking great. I have one question though, would there be a way to 'ignore' certain custom blocks in different Webpack builds? Perhaps this is just me missing something in the documentation, or maybe it's something that needs to be added but, as an example, I wouldn't necessarily want to have vue-loader parsing documentation or unit-tests when I'm in the middle of development. 

I guess maybe with Webpack's caching system this might not make a huge performance issue, but even when thinking about CI systems, it would be nice to be able to have some way to "ignore" certain custom blocks.

asselin:
@iamkether If you want to leave some part of the .vue file out of the generated code, you can specify to use the null-loader in the webpack config file (for example, omit unit tests from debug & release builds).  vue-loader will still process that section of the .vue file, but passing it to the null loader will effectively make it cease processing there.

asselin:
@yyx990803 or @kazupon: This PR is ready for a 2nd review (and merge, hopefully).

dokether:
@asselin ah, okay, that makes sense, thank you.

iamkether:
@asselin ah, okay, that makes sense, thank you.

docs/en/start/spec.md
```diff
@@ -22,6 +22,10 @@ export default {
   color: red;
 }
 </style>
+
+<custom1>
+	This could be e.g. documentation for the component.
```
kazupon: Do we need indent?

docs/en/start/spec.md
```diff
@@ -66,6 +70,66 @@ More details can be found in [Using Pre-Processors](../configurations/pre-proces
 
 - By default, contents will be extracted and dynamically inserted into the document's `<head>` as an actual `<style>` tag using `style-loader`. It's also possible to [configure Webpack so that all styles in all components are extracted into a single CSS file](../configurations/extract-css.md).
 
+### Custom Blocks
+
+Additional custom blocks can be included in a `*.vue` file for any project specific needs.  `vue-loader` will use the tag name to look up which webpack loaders should be applied to the contents of the section.  The webpack loaders should be specified in the `loaders` hash of the `vue` section of the webpack configuration in the same way that languages are specified for the standard sections of the file.  See [Advanced Loader Configuration](../configurations/advanced.md).
```
kazupon: Please arrange the space.

lib/loader.js
kazupon: Did you code updated?

asselin: Sorry, should have replied to this.  This won't throw because there's a guard on line 289 that emits a warning in this case.  This is covered by the `warning should be raised when a loader is not specified for a custom block` unit test.

docs/en/start/spec.md
asselin: It was indented by a tab.  Changed to 2 spaces.

docs/en/start/spec.md
asselin: Sorry, I'm not clear on what you're asking for here.  Is it 1 space between sentences instead of 2?  If so, I just changed that and will push.

asselin:
@kazupon I made updates to the documentation, and see my comment about the test not being necessary.  If I didn't make the changes you were looking for on the doc, please let me know.  Thanks!

docs/en/start/spec.md
kazupon: sorry, my poor explanation. :bow:
exactly!

kazupon:
@asselin Thanks! LGTM!

/ping @yyx990803
Please check just in case.

asselin:
@yyx990803 Is this good to merge now?  Any other cleanup needed?

FYI, I have a follow-on PR that I'm going to submit that adds support for adding attribute values from custom blocks to the webpack command as query parms.

Thanks!

yyx990803:
@asselin sorry for dragging on, just busy with other stuff and don't want to rush-review this. Will try to finalize this week or next.

asselin:
Rebased in on current master, and added a fix for extract-text-plugin 2.0 rc

asselin:
@yyx990803 I see that you're spending some time on vue-loader; any chance this PR could be merged soon?

Issue #497, opened 2016-11-25
override defaultLang


cj:
It would be nice to be able to override https://github.com/vuejs/vue-loader/blob/master/lib/loader.js#L30-L34 in the vue settings.

JounQin:
@yyx990803 I've tried to implement it in pr#500

yyx990803:
Closing outdated issues - if your issue still persists in the latest version of `vue-loader`, please file new one using [the issue helper](http://new-issue.vuejs.org/?repo=vuejs/vue-loader), thank you!

PR #500, opened 2016-11-26
implement a feature of `defaultLang`(#497)

JounQin:


blake-newman:
Hmm I think specifying Lang attributes is useful, especially for large projects.

A new starter to a project may not be aware of the toolset in use. So specifying the Lang attribute in the code can help them navigate to a correct resource.

Only a minor thought.

JounQin:
@blake-newman then just don't use this feature, it's not required.

barraponto:
i don't understand why you need a default language option.
right now you can explicitly tell what language you're using in the template, and also how to deal with any "language". notice that i can make up any value as i need: i can use `<style lang="australian-css">` and then configure vue-loader to pass `australian-css` through a postcss configuration.

JounQin:
If I'm using stylus in every *.vue* file, with this feature we don't need to write `lang="styl"` everywhere, same with `pug` as temple.

Please notice that, it's not required, if you don't like it, you can just don't use it.

JounQin:
Of course, you also could overwrite the language with the original syntax `<style lang="australian-css">`.

barraponto:
In your _stylus everywhere_ scenario, wouldn't you be able to set stylus-loader as the default loader for css just by setting the vue-loader options? As in http://vue-loader.vuejs.org/en/configurations/advanced.html

```
module.exports = {
  // other options...
  module: {
    // module.rules is the same as module.loaders in 1.x
    rules: [
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          // `loaders` will overwrite the default loaders.
          // The following config will cause all <style> tags without "lang"
          // attribute to be loaded with stylus-loader
          loaders: {
            css: 'stylus-loader'
          },
        }
      }
    ]
  }
}
```

JounQin:
The option `loaders` is not designed for it, and if the default pre-processor is incompatible with `css` syntax (for example `sass` with ident syntax), then in you solution, we will not be able to override the lang(`<style lang="css"></style>`) the use plain css anymore.

But with the `defaultLang` feature, you still can use plain css syntax with sass.

octref:
Please don't.

This makes it impossible to write textmate grammar that syntax-highlights embedded code correctly.

All users of Sublime / Atom / VSCode will be affected.

JounQin:
@octref OMG, it is just an optional feature, OK? You don't need it? Just don't use it.

octref:
I agree with @blake-newman in that this adds implicitness which makes code less readable. As scss/less/postcss are syntactically similar, with this feature you can write them all in `<style>`, and without going to the setting you wouldn't know which one is being used.

This also strongly couples code with config, so a piece of vue SFC code can't be used in another project if they don't have the same setting.

And as @barraponto pointed out, if you really want to do it you can change the config yourself. Just beaware the syntax highlighting will break, and it's impossible to change the grammar to support this feature.

barraponto:
> octref OMG, it is just an optional feature, OK? You don't need it? Just don't use it.

I guess that's not the point. It is, of course, up to the maintainer to choose whether to merge or not your PR. Yet features do not come for free -- they require tests, documentation and maintenance. And as this feature has its own effects on code legibility, it's worth weighing them out.

Open source means you can add all the features you want to your versions (forks) of any piece of code. Yet it doesn't mean every PR should be merged upstream. Try not to take the outcome of this PR personally.

PR #532, opened 2016-12-12
[Feature]loader option accepts function

xjchengo:
This PR aims at addressing the problem of #531.


kazupon:
Thank you for your PR!
Sorry for the late reply.

Please resolve the conflict.

kazupon:
> Please resolve the conflict.

never mind, I fixed it.

xjchengo:
Sometimes, we need to overwrite default html loaders. But in webpack config, we cannot access to module id. So if default html loaders are overwritten, hot-reload feature will miss.

yyx990803:
Closing due to inactivity - we are also in feature freeze for v14 and below. Thanks for the PR.

Issue #547, opened 2016-12-21
npm test throws error:  Uncaught TypeError: _vm._v is not a function


airyland:

btipling:
I'm seeing this same error in my hello world vue.js app that worked prior to upgrading. I upgraded from `"vue": "^2.1.3"` to `"vue": "^2.1.6"` this should not be a breaking upgrade.

btipling:
FWI I have seen this be a problem when I have a mismatch between the version used for building vue and the one it's being run with. For example I upgraded my local vue.js to the most recent version and then deployed that built version with to an image with an old version of vue.js and saw this exact same error. Making sure the image had the right version of vue.js fixed this for me.

asselin:
I'm getting a similar error,  Cannot read property '_c' of undefined.  As far as I can tell, test.js's mockRender function doesn't have the required properties.  For the case I reported, the code that blows up is `var _c=_vm._self._c||_h` (there is no _self passed in from test.js).  For the _v issue, later there is code `[_vm._v(_vm._s(_vm.msg))]`, and _v isn't passed in either.  This appears to be generated by vue-template-es2015-compiler's buble.js.

asselin:
The maximum versions of compatible dependent packages supported are:

vue-template-es2015-compiler: 1.3.1
vue: 2.1.4
vue-template-compiler: 2.1.4

Anything after those versions makes npm test throw errors due to changes to the compiler that haven't been propagated to test.js's `mockRender()`

airyland:
@asselin 👍

kazupon:
Close (#550 merged)

PR #550, opened 2016-12-25
Fix mockRender to work with the changes in compiler API introduced in…

asselin:
… vue/vue-template-compiler 2.1.5 and vue-template-es2015-compiler 1.3.2


asselin:
Fixes issue #547

asselin:
I think circleci is pulling in old npm modules for some reason, thus the tests fail.

kazupon:
Thanks!
I rebuild without cache.

Issue #569, opened 2017-01-10
Loading an image source within SVG <image> xlink:href fails


TheLarkInn:
(Hi all working on vue-loader, we came across this issue in our core repo from @sdras, moving it here:)

First off, thanks so much for your hard work on webpack, I use it every day! 🙌

After speaking to @TheLarkInn, I do believe that this is a bug, though likely low use-case. (until people figure out how awesome this implementation is 😎)

Currently if I load an image with a relative src within a .vue file, it works, but images embedded within an SVG DOM are not recognized. This is probably due to the XML markup for image, where you would create an  tag and then within it use an xlink:href="" attribute instead of src="" to load the relative path. When I use an absolute path, I observe the correct behavior, and the image loads.

Benefits of using the SVG to load an image include but are not limited to:

image positioning: for dynamic content, an SVG image can be placed exactly without absolute positioning because you're using the SVG coordinate system
responsive, as this positioning will be stable along with other images within the same coordinate system
I'm on a MacBook Pro, node version v6.7.0.

Thank you very much for your attention in the matter!

sdras:
thank you @TheLarkInn!

TheLarkInn:
@yyx990803 let me know if we can help with this I haven't spent enough time looking in vue-loader as of late. No prob @sdras :)

yyx990803:
You should be able to use `vue-loader`'s [transformToRequire](http://vue-loader.vuejs.org/en/options.html#transformtorequire) option to make that work:

``` js
// pass this to vue-loader
// see http://vue-loader.vuejs.org/en/options.html
{
  transformToRequire: {
    image: 'xlink:href'
  }
}
```

We may consider adding this to default config.

sndrgb:
Guys form me transformToRequire is ok for normal binding but if i want to use v-bind:src with me it doesn't function, normal?

LinusBorg:
Yes, normal. I you use b-bind, the attribute's value is a Javascript expression, not a source path. vue-loader can't somehow guess the path from that.

You have to require the image yourself in the Javascript code of your component.

Yearhouse:
Im trying this but i cant get the svg paths to change:

`{ 
              test: /\.vue$/,
              loader: 'vue-loader',
              html: {
                  root: path.resolve(__dirname, '/static/production/assets')
               },
              options: {
          		transformToRequire: {
          			image: 'xlink:href'
          		}
          	}`

im my Vue template file
[...]

PR #579, opened 2017-01-17
Add 'image' tag of SVG to 'transformToRequire` option as default

kazupon:
ref: https://github.com/vuejs/vue-loader/issues/569#issuecomment-271710412

@yyx990803 said 💬 
>We may consider adding this to default config.

@vuejs/vue-loader-vueify 
Please review ! 👀

### DIFF
--- a/docs/en/options.md
+++ b/docs/en/options.md
@@ -108,9 +108,9 @@ module.exports = {
 ### transformToRequire
 
 - type: `{ [tag: string]: string | Array<string> }`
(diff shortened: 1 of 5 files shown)


yyx990803:
👍

blake-newman:
@yyx990803 @kazupan

SVG also have the use tag that could also be added by default. This is used for SVG spritemaps.

Although with http 2 this becomes less relevant I still use this for icons ect.

PR #582, opened 2017-01-19
Refactor Vue Loader to be usable outside of WebPack

paulcbetts:
This is the first of a series of PRs to try to get vue-loader working in contexts other than WebPack. Specifically, I want to try to add support for Vue 2.0 components into https://github.com/electron/electron-compile, which is an Electron library used to transparently add support for lots of web langauges to Electron apps without any build steps.

This PR has no functional changes whatsoever other than exporting a new class `VueLoader` which exposes some of the guts of vue-loader. It also lowers the reliance of the code on global variables / WebPack-specific internals, and lets the outer method set that up.


yyx990803:
Not sure if you are aware of it, but the way `vue-loader` works is by loading the different parts of the same `*.vue` file using different inline request strings (e.g. `require('a-loader!b-loader!./some-file.vue')`) - which is a Webpack-specific mechanism. It won't work without Webpack unless `electron-compile` supports the same inlinie request, or you plan to handle this on your own.

paulcbetts:
@yyx990803 That's next on my list :)

yyx990803:
@paulcbetts if that's the case, I'm not sure if the refactoring should happen inside `vue-loader`'s codebase - the more ideal solution would be a standalone dependency: an abstract vue component compiler that exposes an API so it can be shared across `vue-loader`, `vueify` and `electron-compile`.

paulcbetts:
@yyx990803 That seems ambitious to try to accomplish as an external contributor

yyx990803:
@paulcbetts I know, but imo it's even more ambitious to rewrite `vue-loader` to achieve the same because working inside `vue-loader` also implies the liability to ensure everything that depends on `vue-loader` continues to work properly. It seems less risky to me to go with a separate proof-of-concept and then modify it to work with `vue-loader` - in fact I'd say the codebase of `vueify` is more suitable to fork from because it relies less on Webpack's loader mechanisms and is more straightforward string compilation.

paulcbetts:
@yyx990803 I don't actually think this to be the case - I really only have to replace the style `require` and the script `require` with a custom version, which we can easily make injectable as an optional method. 

Since electron-compile already handles 90% of the script transform already (the only thing it doesn't is the "select script from vue template" bit), it's closer than you think!

paulcbetts:
This PR is completely merge conflicted now, which is exactly what I was afraid of happening once it stalled. Let's just forget about it.

yyx990803:
Sorry for the wasted effort - however I think this repo is simply to webpack-specific to generalize and it's better to keep it that way to reduce maintenance burdens. I still think forking off vueify would be a better option here.

michael-ciniawsky:
@paulcbetts Maybe take a look at [PostHTML](https://github.com/posthtml/posthtml) for a standalone 'Single File Component' Compiler (Plugin)

```js
module.exports = (options) => {
  return function posthtmlComponent (ast) => {
     ast.walk((node) => {
         if (node.tag === 'style') doSomethingWithStyle(node)
         if (node.tag =='template') doSomethingWithHTML(node)
         if (node.tag === 'script') doSomethingWithScript(node)

         return node
     })

    return ast
  }
}
```

paulcbetts:
@michael-ciniawsky This does look promising, thanks for the tip

PR #586, opened 2017-01-20
Use assign

anteriovieira:


yyx990803:
Thanks for the PR - since we now assume minimum version of Node 4+, we will just use `Object.assign` directly from now on.

PR #607, opened 2017-01-30
add query parameters chaining section

kazupon:
This feature docs is supported with the below commits:
- https://github.com/vuejs/vue-loader/commit/e2b41eac5528152396a703a594aa2ef4af5216c4 (for Vue 2.x)
- https://github.com/vuejs/vue-loader/commit/39082d4fc3d5181dd90c6e9e790a69cd878e1cc9 (for Vue 1.x)

### DIFF
--- a/docs/en/configurations/chaining.md
+++ b/docs/en/configurations/chaining.md
@@ -1,0 +1,43 @@
+# Query parameters chaining
+
+> requires ^10.1.0 for Vue 2.x, ^8.6.0 for Vue 1.x
+
+Sometimes you may want to bypass a your custom loader, before pass to `vue-loader`. To do that, you can customize with the `loader` option.
+
+below example use the your custom loader:
+
+### Webpack 1.x
+
+``` js
+// webpack.config.js
+module.exports = {
+  // other options...
+  module: {
+    loaders: [
+      {
+        test: /\.vue$/,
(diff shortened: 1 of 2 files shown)


docs/en/configurations/chaining.md
```diff
@@ -0,0 +1,43 @@
+# Query parameters chaining
```
yyx990803: I'd just call this "Chaining with Other Loaders" instead of parameters

docs/en/configurations/chaining.md
```diff
@@ -0,0 +1,43 @@
...
+
+> requires ^10.1.0 for Vue 2.x, ^8.6.0 for Vue 1.x
+
+Sometimes you may want to bypass a your custom loader, before pass to `vue-loader`. To do that, you can customize with the `loader` option.
```
yyx990803: Let's change this to "you may want to pre-process the vue file before passing it to `vue-loader`"

kazupon:
Thanks for your reviews! 👀 
I fixed and pushed it! ✨

yyx990803:
Hmm actually after looking at this, I don't think we need to explicitly document this. Most loaders support chaining like this which is just part of how Webpack works - this is just supporting something we should've supported in the beginning.

Thanks for the suggestions though!

kazupon:
Indeed 🤔
It's okay 😺

Issue #592, opened 2017-01-24
CSS Module class names have a case where they don't hot reload within templates


trainiac:
There is one aspect of my vue component that does not hot reload. Load the example vue component below in a page with all the webpack hot reloading business configured.

```
<template>
  <div :class='$style.foo'>
    Hello
  </div>
</template>

<script>
  export default {
    name: 'App'
  }
</script>

<style module>
  .foo {
    color: red;
  }
</style>
```

Then once page has loaded change $style.foo -> $style.fo and change the corresponding css class name .foo -> .fo.  I would expect the module would reload and there would be no change in the page appearance.  What I noticed after inspecting the DOM, is that while the styles get regenerated correctly the template class attribute is now empty as if $style.fo was not defined in the template.

Examples where hot reloading _does_ work:

1. If I load the same component as above with .fo already defined as a css class, then change $style.foo -> $style.fo in the template, then save, the div will pick up the styles of .fo.  

2. If I loaded the same component as above, create a new css class .fo and a new element <span v-bind:class='$style.fo'> in the template referencing the new class, then save, the component will hot reload as expected.
  
It seems the problem is that an element bound to a css module class name once loaded, cannot reference newly created css module classes without having to refresh the entire page.

I can create a simple test case repo to demonstrate, but wanted to know if anyone else had confirmed this behavior before spending more time on this.

trainiac:
I have created a repository where the configuration is as minimal as possible. That demonstrates the aforementioned bug.
https://github.com/trainiac/VueLoaderCSSModuleNameBug

trainiac:
@kazupon let me know if you need more information. I'll try to debug if I have time this weekend.

kazupon:
Thank you for your feedback!
Hmm, I tried to reproduce this issue, however I could not check it. 🤔

trainiac:
Hmmm. So you were not able to reproduce?

Just to confirm, steps to reproduce:

1. Load the page in a browser
2. Change the template $style.foo -> $style.fo and the style .foo -> .fo
3. Save
4. See the browser reloads but the div text is no longer red because it doesn't have a class on it anymore.

kazupon:
Thank you for your reproduction steps!
I could reproduce this issue.
This is bug. 🐛

trainiac:
It's a tough bug to find because it most cases for me everything else is working perfectly! Thank you.

kazupon:
fixed

PR #609, opened 2017-01-30
fix cannot set css modules to vue instance

kazupon:
Reference (Reproduction repo too): #592 

In css modules, When we change styles, these styles can not set to Vue Instance due to cache of Vue Constructor.
I fixed this issue with using HMR API.


yyx990803:
👍

PR #655, opened 2017-02-18
Change occurence to occurrence

oburatongoi:
This was causing the following error:

webpack.optimize.OccurenceOrderPlugin is not a constructor

### DIFF
--- a/docs/en/workflow/production.md
+++ b/docs/en/workflow/production.md
@@ -24,8 +24,8 @@ module.exports = {
         warnings: false
       }
     }),
-    // optimize module ids by occurence count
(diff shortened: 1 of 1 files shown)


Alex-Sokolov:
Maybe also add comment that [OccurrenceOrderPlugin not needed with webpack 2](https://webpack.js.org/guides/migrating/#occurrenceorderplugin-is-now-on-by-default)?

kazupon:
Yeah, Indeed.
I'll revert.

kazupon:
Oh, wait.
vue-loader are support webpack 1.x and 2.x
I think we should be commented for webpack 1.x 📝

PR #664, opened 2017-02-21
feat: Call custom block loaders with `Component` as argument

hauptbenutzer:
I love the idea of single file vue components and of being able to add custom blocks such as `<docs>`. Although it is possible to pass arguments to the loader via attributes on the custom block tag, you cannot access the `Component` from inside the loader. I needed that functionality to achieve the following. 

1. Have a `<docs>` tag in the .vue file 
2. Have the contents of said tag inserted into the resulting Component as `options._docs` so that it would be accessible during runtime. 

A very simple custom loader for this task could, for example, look like this: 

```javascript
module.exports = function (source, map) {
  this.callback(null, 'module.exports = function(Component) {Component.options._docs = ' +
    JSON.stringify(source) +
    '}', map)
}
```

So I changed the call of the customBlock to include `Component`, if the result of the `require` is a function. I'm not sure if this would break any existing workflows. 

If you think this use case is worth considering I'd be happy to supply the corresponding unit tests and docs.


yyx990803:
Neat idea, mind adding a test case for this?

hauptbenutzer:
Sure thing! I've added some example docs as well.

test/mock-loaders/docs.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function (source, map) {
+  this.callback(null, 'module.exports = function(Component) {Component.options.__docs = ' +
+    JSON.stringify(source) +
+    '}', map)
+}
```
phanan: Nitpicky me here: I think we need a new line.

yyx990803:
@hauptbenutzer thanks for the PR and great docs example!

Issue #666, opened 2017-02-21
DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic


cfv1984:
Hi all! I made a new webpack-simple project, and when running npm run build I get presented with the message in the title. Specifically: 

```cli
404s will fallback to /index.html
webpack: wait until bundle finished: /dist/build.js
(node:568) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```

This is visible when using ```performance: { hints: "warning" }``` in your webpack config.

Additionally, doing an npm run dev I get this warning text: 

```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (250 kB).
This can impact web performance.
Assets: 
  build.js (1.55 MB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (250 kB). This can impact web performance.
Entrypoints:
  main (1.55 MB)
      build.js


WARNING in webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/

```

What can be done about any of this two things?  This is coming from a plain new ```vue init webpack-simple $myproject``` build, not something particularly involved and I'm not sure it is intentional

lanlin:
same warning

![image](https://cloud.githubusercontent.com/assets/3468024/23193721/510a752c-f8e5-11e6-918c-7913ccd6d1cc.png)

here the link [https://github.com/webpack/loader-utils/issues/56](https://github.com/webpack/loader-utils/issues/56)

s20081428:
same here

shmy:
same problem here

edmondyip:
How can i fix it?

rainjay:
set `process.noDeprecation = true` , the warning will be removed

roscoe054:
@kazupon The warning seems to reappear at the line:
 https://github.com/vuejs/vue-loader/blame/master/lib/loader.js#L146

Been101:
where is the webpack config? how to set process.noDeprecation = true?

miyalee:
@rainjay where? use vue-cli,can't find any file named"webpack.config.js", "webpack.base.conf.js"? how to add it?

beckyao:
how to fix?

[...]

PR #672, opened 2017-02-23
fix deprecated warning for 8.x

kazupon:
related: #666


gbezyuk:
nice related issue id, huh )

DeepenLau:
It seems that this problem is not release to npm.
After i run npm i vue-loader@8 -D , and the version is still the not fix version.
Thanks~

image


@yyx990803

kokororin:
@yyx990803 Can you publish fixed version for 8.x ?

Issue #675, opened 2017-02-27
yarn update issue


steambap:
When I update vue-loader to the latest (11.1.3), yarn package manager will use vue-hot-reload-api version 2.0.6, not the latest(2.0.11). And it refuse to use the latest vue-hot-reload-api even if I explicitly add it as dependency. This will result in runtime error with latest vue esm runtime. 

I know this is probably yarn's issue, but I assume that this can be resolved simply by bump dependencies in package.json?

rferro:
+1

jialanmo:
+1

blade254353074:
+1

sqal:
Yeah, this causing error in my App since I updated to vue@2.2.1 vue-router@2.3.0

![img1](http://image.prntscr.com/image/09b944b89e924a3f93c997c17d548065.png)
![img2](http://image.prntscr.com/image/d6b568a40afe45909628f55196953c4a.png)

Had to disable HMR for now...

blade254353074:
I made a pr to update the dependencies.
Waiting for

LinusBorg:
I assume this is because yarn caches dependencies on your machine for some time.

you could clean the cache with `yarn cache clean`

sqal:
@LinusBorg ok I managed to resolve my previus issue by cleaning yarn cache, removing node_modules/yarn.lock and installing dependencies once again. But if i may, I have one more question though. I have disabled production tip in my app via config `Vue.config.productionTip = false` but this only works when HMR is disabled. Why is that?

LinusBorg:
Sorry to say, but I have no idea :D It may have something to do that the HMR API loads vue itself during the replacement process, in somehwere in that proceess the setting gets lost? 

Maybe open an issue?

yyx990803:
published 11.1.4 with bumped `vue-hot-reload-api` version.

PR #677, opened 2017-02-27
update dependencies

blade254353074:
When using yarn to install dependencies, yarn will lock the vue-hot-reload-api version to 2.0.1.
After update vue to 2.2.1...

image

So I updated the vue-loader dependencies ,run the test, and made this pr.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "postcss-load-config": "^1.1.0",
     "postcss-selector-parser": "^2.0.0",
     "source-map": "^0.5.6",
-    "vue-hot-reload-api": "^2.0.1",
+    "vue-hot-reload-api": "^2.0.11",
     "vue-style-loader": "^2.0.0",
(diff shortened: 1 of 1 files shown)


xalexec:
+1

kazupon:
related #675

LinusBorg:
I don't think that all these dependency version bumps are necessary or related to this problem, and in case of inject-loader, we csn't use 3.0.0 because it doesn'T work correctly.


Can you limit the update to those packages that are problematic when updating with yarn?

blade254353074:
@LinusBorg My bad. I was too rash to update the dependencies...
Now I just update vue-hot-reload-api to 2.0.11.

LinusBorg:
/ping @yyx990803 

A lot of people had problems getting this dependency to update correctly with yarn, even with deleting `/node_modules`, I think this should solve it.

PR #679, opened 2017-02-28
Use ESM build with require

znck:


znck:
Closing because `vue-hot-reload-api` has interop built-in.

PR #690, opened 2017-03-06
add missing comma

guylyons:
tiny fix, just missing comma in JSON example


guylyons:
np! i've updated ru and pt_BR, but kr appears to be out of date and did not contain those lines of json.

kazupon:
Thanks!

PR #702, opened 2017-03-09
when mismatch version, it is better to fail the loader and webpack show error code

sunlei33:
instead of checking version in vue-template-compiler runtime, check the version in vue-loader.
when mismatch version, the loader will fail and the webpack will return the error code.

### DIFF
--- a/lib/template-compiler.js
+++ b/lib/template-compiler.js
@@ -58,6 +58,13 @@ module.exports = function (html) {
       options.transformToRequire
     )
   }
+  
+  var compilerVersion = require('vue-template-compiler/package.json').version
+  var vueVersion = require('vue').version
+
+  if(vueVersion !== compilerVersion){
+    this.emitError(`vue version ${vueVersion} mismatch vue-template-compiler version ${compilerVersion}`)
+  }
 
   var compiled = compiler.compile(html, Object.assign({
     preserveWhitespace: options.preserveWhitespace
--- a/lib/template-compiler.js
+++ b/lib/template-compiler.js
@@ -58,11 +58,10 @@ module.exports = function (html) {
       options.transformToRequire
     )
   }
-  
   var compilerVersion = require('vue-template-compiler/package.json').version
   var vueVersion = require('vue').version
 
-  if(vueVersion !== compilerVersion){
+  if (vueVersion !== compilerVersion) {
     this.emitError(`vue version ${vueVersion} mismatch vue-template-compiler version ${compilerVersion}`)
   }
 
--- a/lib/template-compiler.js
+++ b/lib/template-compiler.js
@@ -58,11 +58,19 @@ module.exports = function (html) {
       options.transformToRequire
     )
   }
-  var compilerVersion = require('vue-template-compiler/package.json').version
-  var vueVersion = require('vue').version
 
-  if (vueVersion !== compilerVersion) {
-    this.emitError(`vue version ${vueVersion} mismatch vue-template-compiler version ${compilerVersion}`)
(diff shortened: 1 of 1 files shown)


lib/template-compiler.js
```diff
@@ -58,6 +58,12 @@ module.exports = function (html) {
       options.transformToRequire
     )
   }
+  var compilerVersion = require('vue-template-compiler/package.json').version
+  var vueVersion = require('vue').version
```
yyx990803: probably use `vue/package.json` here too to avoid evaluating Vue itself on startup.

lib/template-compiler.js
```diff
@@ -58,6 +58,12 @@ module.exports = function (html) {
...
+
+  if (vueVersion !== compilerVersion) {
+    this.emitError(`vue version ${vueVersion} mismatch vue-template-compiler version ${compilerVersion}`)
```
yyx990803: Let's use the same language found [here](https://github.com/vuejs/vue/blob/dev/packages/vue-template-compiler/index.js#L9-L14)

lib/template-compiler.js
sunlei33: commit a patch for this PR

lib/template-compiler.js
sunlei33: commit a patch for this PR

yyx990803:
After testing this locally, I'm actually not sure if this is better - this change requires removing the version check in `vue-template-compiler`, but that also means when it is used in other contexts other than `vue-loader` it would no longer enforce correct versions.

Either way, the user would need to install the correct versions.

alamin3000:
@yyx990803 How do you expect us to resolve this? I change the vue version that the template compiler was complaining about to 2.2.4, now vue-server-renderer is complaining that it needs the 2.2.6 back.... seriously.... you need to reconsider the patch or some other solution.  Thanks.

yyx990803:
@alamin3000 if you don't know how to pin versions, just do a fresh install.

alamin3000:
@yyx990803 Thanks.  But I wish everything was as easy as "restart the computer" or reinstalling something.  Basically I am trying to create a node module (separate repo from my main app) for Vue based template parser say, vue-template-parser.  So it has its own Vue dependency + vue-server-renderer.  And the app that will use that library also has vue as its dependencies.  I can't make sure/dictate that whoever uses the parser they also need to have EXACT same version of the vue... like 2.4.1 (including the patch version number).  At least if it ignored the minor patch numbers I would be little less frustrated.  Anyway, hope there is a better solution for this or I'm doing something wrong so I can fix it.

yyx990803:
@alamin3000 In that case your module should list Vue and vue-template -compiler as peer deps

alamin3000:
@yyx990803 I do that.  The module works independently.  It compiles and renders strings.  But issue happens when I try to use that module which already has vue as dependency!  Then I need to make sure that my app vue matches EXACTLY with the module's vue version. One of the advantage of NPM over others such as bower/ composer (php) is that every module can have same library in with different version.  If the loader forces otherwise it breaks the way npm module works.

alamin3000:
Here is the repo.  https://github.com/es-compose/es-renderer-vue that uses vue for parsing html. If I use this library in any project that has other vue version (2.4.1) then it will complain.

yyx990803:
You are listing them as **dependencies**, not **peerDependencies**.

alamin3000:
peerDependencies what now!?# I didn't know it existed.  Okay I will read on it and try to implement that. Thanks @yyx990803

PR #729, opened 2017-03-23
fix parse cssModules loader query

guox191:
Hi,
I think it should use `loaderUtils.parseQuery` to parse the result from regular expression.
Otherwise we can't get css-loader' query such like `...!css-loader?{"minimize":true}...` for css-module.
The last change RP: https://github.com/vuejs/vue-loader/commit/1d63e9ec610490767b447cee5cf53e00faf905fa
Thanks

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -145,7 +145,7 @@ module.exports = function (content) {
     return loader.replace(/((?:^|!)css(?:-loader)?)(\?[^!]*)?/, function (m, $1, $2) {
       // $1: !css-loader
       // $2: ?a=b
-      var query = loaderUtils.getOptions($2 || {}) || {}
+      var query = loaderUtils.parseQuery($2 || '?')
       Object.assign(query, OPTIONS, option, DEFAULT_OPTIONS)
       if (index !== -1) {
         // Note:


guox191:
It seems circleci use old `loader-utils` version which didn't export `parseQuery` method

lib/loader.js
```diff
@@ -145,7 +145,7 @@ module.exports = function (content) {
     return loader.replace(/((?:^|!)css(?:-loader)?)(\?[^!]*)?/, function (m, $1, $2) {
       // $1: !css-loader
       // $2: ?a=b
-      var query = loaderUtils.getOptions($2 || {}) || {}
+      var query = loaderUtils.parseQuery($2 || '?')
```
znck: `parseQuery` is deprecated.

guox191:
@znck 
hello
This is latest `loader-utils` source code: https://github.com/webpack/loader-utils/blob/master/lib/getOptions.js
```
function getOptions(loaderContext) {
	const query = loaderContext.query;
	if(typeof query === "string" && query !== "") {
		return parseQuery(loaderContext.query);
	}
	if(!query || typeof query !== "object") {
		// Not object-like queries are not supported.
		return null;
	}
	return query;
}
```
`getOptions` receive a loader context object, not a string like now.

And `parseQuery` is not deprecated in latest version, see: https://github.com/webpack/loader-utils/commit/ddf64e49c5871c7229a5a536656a711479c08516

rguanghui:
vue-loader@10.2.0 not compress css now when i use css moules.

znck:
@guox191 My bad. Thanks for PR.

Issue #759, opened 2017-04-05
Warnings thrown when using Rewire


matt-oconnell:
Getting `"named exports are not supported in *.vue files."` error using newest version of `vue-loader`

We are using [The rewire babel plugin](https://github.com/speedskater/babel-plugin-rewire) which is causing [this line](https://github.com/vuejs/vue-loader/blame/3878a62a712f5bbbc29de9dbd928b130119b39e9/lib/loader.js#L392) to fail within `vue-loader`. The issue seems to be that the check within `vue-loader` is looking for *any* properties other than the `default` and `__esModule` property. Rewire inserts some extra "private" properties on the component:

![screen shot 2017-04-05 at 13 03 13](https://cloud.githubusercontent.com/assets/8006041/24717468/9c321f14-1a00-11e7-9477-4491e6b6ebf9.png)
 
A proposed fix would be to check for any "non-private" properties, while still allowing objects with properties prefixed with `__` to pass this check not log an error. This may or may not be considered the responsibility of the `vue-loader` project but if it is, let me know and I'd be happy to make a PR

yyx990803:
Hmm this is a bit niche but I guess we can take a PR for that.

matt-oconnell:
Great, I will try and get to it within the next week or two.

PR #765, opened 2017-04-08
Do not consider private properties to be named exports

matt-oconnell:
fix: https://github.com/vuejs/vue-loader/issues/759


matt-oconnell:
@yyx990803 Any feedback on this?

PR #797, opened 2017-04-28
[Fix] functional SFC style injection support

blake-newman:
Adds support for SFC components that have injectStyles. 

Example component: 
```
<script>
  export default {
    functional: true,
    render(h, { children }) {
      return h('div', children);
    },
  };
</script>

<style lang="scss">
  @import 'style/settings';
  @import 'style/tools';
  @import 'style/components/wrapper';
</style>
```

Also does not apply hook to functional components, without injectStyles. I believe that this is not needed, correct me if i'm wrong.

### DIFF
--- a/lib/component-normalizer.js
+++ b/lib/component-normalizer.js
@@ -40,7 +40,10 @@ module.exports = function normalizeComponent (
   if (moduleIdentifier) { // server build
     hook = function (context) {
       // 2.3 injection
-      context = context || (this.$vnode && this.$vnode.ssrContext)
+      context =
+        context || // cached call
+        (this.$vnode && this.$vnode.ssrContext) || // stateful
+        (this.parent && this.parent.$vnode && this.parent.$vnode.ssrContext) // functional
       // 2.2 with runInNewContext: true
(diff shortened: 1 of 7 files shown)


lib/component-normalizer.js
```diff
@@ -61,7 +61,16 @@ module.exports = function normalizeComponent (
     hook = injectStyles
   }
 
-  if (hook) {
+  if (hook && options.functional && injectStyles) {
+    // inject styles for functioal component in vue file
+    const existing = options.render
+    options.render = function renderWithStyleInjection (...args) {
```
yyx990803: This piece of code will run in browser without compilation so we need to use ES5

blake-newman:
@yyx990803 Updated to use es5

lib/component-normalizer.js
```diff
@@ -61,7 +61,16 @@ module.exports = function normalizeComponent (
     hook = injectStyles
   }
 
-  if (hook) {
+  if (hook && options.functional && injectStyles) {
+    // inject styles for functioal component in vue file
+    const existing = options.render
```
yyx990803: `const` is still es6 ;)

PR #819, opened 2017-05-18
fix incorrect regexp

jin5354:
fix incorrect regexp.


nickmessing:
That makes some sense 👍

jin5354:
Sure 😄

kazupon:
Thanks!

PR #822, opened 2017-05-21
"npm install -g" requires "root" privileges on Linux

famzah:
Tested on Linux. Alternatively, we can install without the "-g" option which works only for the current user, in which case we need to create a symlink manually:
```bash
ln -s ~/node_modules/vue-cli/bin/vue ~/bin/vue
```

### DIFF
--- a/docs/en/start/setup.md
+++ b/docs/en/start/setup.md
@@ -5,7 +5,7 @@
 It's recommended to scaffold a project using `vue-loader` with `vue-cli`:
 
 ``` bash
-npm install -g vue-cli
+sudo npm install -g vue-cli
 vue init webpack-simple hello-vue
 cd hello-vue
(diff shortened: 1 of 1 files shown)


famzah:
I'm not sure whether this instruction will work on Windows though...

famzah:
The `-g` option operates in "global" mode which installs modules in `/usr`. This always requires `root` privileges.

I don't insist that you add this to the docs and you can close the PR right away, if you still believe that `sudo` will bring more confusion than help.

kazupon:
Thank you for your response!

>The -g option operates in "global" mode which installs modules in /usr. This always requires root privileges.

Yes, In generally, -g option requires root privileges.
I think that user can understand to the environment whether `sudo` is necessary.

PR #823, opened 2017-05-24
Allows babel to run after typescript

Evertt:
I'm not so experienced with making PRs so please let me know if I'm doing it wrong.

Anyhow, I'm using VueJS with typescript in my own projects and I didn't manage to setup my webpack config in such a way that it would also run babel on my vue-files after running typescript. Until I started messing around in the vue-loader itself and figured out that the piece of code that I'm submitting in this PR did work.

Now you may be thinking "but dude, if you can write `ts: 'babel-loader!ts-loader'` in vue-loader then you can also write that in your own config". But no, for some reason webpack then says it can't resolve `'babel-loader!ts-loader'`, even though it can resolve it when I put it in vue-loader.

So yeah, I hope you're willing to accept this PR, so that I don't have to make my project depend on my own fork to make it work. Because that would mean that I'd need to maintain my own fork as well...

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -19,7 +19,7 @@ var styleLoaderPath = normalize.dep('vue-style-loader')
 var hotReloadAPIPath = normalize.dep('vue-hot-reload-api')
 
 // check whether default js loader exists
-var hasTs    = !!tryRequire('ts-loader')
+var hasTs = !!tryRequire('ts-loader')
 var hasBabel = !!tryRequire('babel-loader')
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR. However, I think it's most likely an issue elsewhere that causes webpack to fail to resolve `babel-loader!ts-loader`, and I don't think patching/forking `vue-loader` is the proper way to fix that problem for the long run.

More importantly, configuring TypeScript with `vue-loader` is more complicated than Babel and this default behavior will only fit a small set of use cases, unlike Babel.

PR #826, opened 2017-05-25
Don't mutate this.options (fixes #824)

### LINKED ISSUE #824: Webpack errors if run more than once due to vue-loader mutating config object
### Version
12.1.0

### Reproduction link
[https://github.com/arichardsmith/vue-loader-bug-report](https://github.com/arichardsmith/vue-loader-bug-report)

### Steps to reproduce
- Clone [the repo](https://github.com/arichardsmith/vue-loader-bug-report) and install dependancies.
- Run `gulp dev`
- Edit any file in ./src

### What is expected?
Gulp is watching the src directory and should run the build task without any errors.

Webpack errors on every build after the first with:
```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property '__vueOptions__'.
  ......
```

---
It seems that vue-loader mutates the webpack config object here:
```javascript
// /lib/loader.js line 52
var options = this.options.__vueOptions__ = Object.assign({}, this.options.vue, this.vue, query)
```

This introduces an option that webpack doesn't allow.
It isn't an issue if you are running webpack from the command line, but when you are using the API, the config object gets re-used for each call, giving the error.

Cloning your config object before you pass it to webpack avoids any of these problems:
```javascript
webpack(config, callback) // bad

webpack(Object.assign({}, config), callback) // good
```

arichardsmith:
As mentioned in #824, modifying the options object causes webpack to error if you run it more than once with the same config.

This PR stops `lib/loader.js` from setting `this.options.__vueoptions__` and stops any other file from relying on it. I'm not sure if any other loader relies on this key, anything that does would be broken by this PR.

I've also had to rely on checking `this._compiler.constructor.name` in `lib/style-compiler/index.js` to check for happypack, I'm pretty sure that's what the original line was doing? It doesn't feel good but it will work. I don't have any experience with happypack, so I can't confirm if this makes a difference.

It passes all tests, I didn't think the changes needed a new test but I can write one if needed.
Hopefully I havn't missed anything!

### DIFF
--- a/lib/style-compiler/index.js
+++ b/lib/style-compiler/index.js
@@ -9,10 +9,10 @@ module.exports = function (css, map) {
   this.cacheable()
   var cb = this.async()
 
-  var query = loaderUtils.getOptions(this) || {}
-  var vueOptions = this.options.__vueOptions__
+  var query = Object.assign({}, loaderUtils.getOptions(this), this.options.vue, this.vue)
 
-  if (!vueOptions) {
+  // Slightly messy check for HappyPack as we can't use this.__vueoptions__ anymore
+  if (this._compiler.constructor.name === 'HappyFakeCompiler') {
     if (query.hasInlineConfig) {
       this.emitError(
         `\n  [vue-loader] It seems you are using HappyPack with inline postcss ` +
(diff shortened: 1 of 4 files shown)


lib/style-compiler/index.js
```diff
@@ -9,10 +9,10 @@ module.exports = function (css, map) {
   this.cacheable()
   var cb = this.async()
 
-  var query = loaderUtils.getOptions(this) || {}
-  var vueOptions = this.options.__vueOptions__
+  var query = Object.assign({}, loaderUtils.getOptions(this), this.options.vue, this.vue)
 
-  if (!vueOptions) {
+  // Slightly messy check for HappyPack as we can't use this.options.__vueoptions__ anymore
+  if (this._compiler.constructor.name === 'HappyFakeCompiler') {
```
kazupon: Hmm, I cannot review this line due to I'm not very familiar with `HappyPack`. 🤔 
/ping @yyx990803

yyx990803:
Thanks for the PR but the fix would actually break when there are non-stringifiable options being passed to the vue options. We should probably update the tests to ensure that's covered.

arichardsmith:
That figures. Thanks for fixing it anyway!

PR #834, opened 2017-05-31
Upgrade PostCSS to 6.0.1

termosa:
PostCSS plugins (like `postcss-nested` in my case) starts requiring a higher version of PostCSS.


yar-usenko:
Please update postcss!

PR #847, opened 2017-06-18
[feature] A smart way to support markdown template language

Dafrok:
## Problem

I met a critical problem when I followed #595 to compile markdown template in vue components.
While compiling markdown template with more than single line, the component template won't contain a root element, and the markdown specification doesn't allowed us to add a wrapper.

### Input

```vue
<-- Case 1 -->
<template>
# Hello world!
## FooBar
</template>

<-- Case 2 -->
<template>
<div>
# Hello world!
## FooBar
</div>
</template>
```

### Expect Output

```html
<div>
  <h1>Hello world!</h1>
  <h2>FooBar</h2>
</div>
```

### Actual Output

```html
<-- Case 1 -->
<h1>Hello world!</h1>
<h2>FooBar</h2>

<-- Case 2 -->
<div>
# Hello world!
## FooBar
</div>
```

## Solution

As far as I know, the dependency ***consolidate.js*** of template compiler has no plan to support markdown syntax (See details: https://github.com/tj/consolidate.js/issues/45). I has added a dependency ***markdown-it*** to treat this case, because other famous markdown parsers (e.g. [marked](https://www.npmjs.com/package/marked)) will overwrite ids of `h1 - h6` elements.

## Usage

I think this is a smart way to use markdown template.

```vue
<template lang="markdown"> <!-- or `lang="md"` as an alias -->
# Foo
## Bar
</template>
```


blade254353074:
If I remember correctly, markdown should support nested (html & markdown) syntax.

Jinjiang:
I think it's too coupled with the template compiler. How about creating a markdown custom block to doing this job? That's more flexible.
Thanks.

Dafrok:
Well, what does the markdown custom block means?
Actually I know that my solution is not very suitable for vue-loader. I've tried to use markdown filter in some template engines (e.g pug), but vue-loader can not detect the changes of markdown files which is imported by template engine. It's a critical problem when using webpack-dev-server and I found no way to treat it.

Dafrok:
I've read the documentation of custom block, i think it's no so convenient for my case, for example:

```markdown
<template lang="md">
# Title

## The first Component

<x-component></x-component>

## The Second Component

<y-component></y-component>

... and so on
</template>

<script>
export default {
  components: {
    XComponent,
    YComponent
  }
}
</script>
```

Jinjiang:
I have tried to create a markdown-loader like:

``` js
const compiler = require('vue-template-compiler')
const marked = require('marked')

module.exports = function (markdown, map) {
  // markdown -> html
  const html = marked(markdown)
  // html -> render functions
  const { render, staticRenderFns } = compiler.compile(`<div>\n${html}</div>`)
  return `
module.exports = function (Component) {
  Component.options.render = function () {${render}}
  Component.options.staticRenderFns = [${staticRenderFns.join(',')}]
}`
}
```

And it worked well with:

``` vue
<markdown>
# {{title}}

## {{subtitle}} - The first Component
</markdown>

<script>
export default {
  data () {
    return {
      title: 'Hello',
      subtitle: 'Welcome to Your Vue.js App'
    }
  }
}
</script>
```

I think that can help you enough.

Thanks.

Dafrok:
Well, I got it. It seems perfect.

Jinjiang:
Additionally I found another way in our test to have a look: https://github.com/vuejs/vue-loader/blob/aaf724d60f8d47e850d79872f6242b01f9f259d3/test/test.js#L648

PR #848, opened 2017-06-19
Correctly read options.minimize

SLaks:


yyx990803:
It is part of the [loader context](https://webpack.js.org/api/loaders/#this-minimize)

PR #864, opened 2017-06-29
use chai syntax for mock test example

LinusBorg:
since we use mocha&chai in our webpack template, it makes sense that the test examples use that syntax as well.


LinusBorg:
Hm, now I see that the "Testing" page is also completely focussed on Jasmine. will incorporate that too later.

PR #866, opened 2017-07-01
Support postcss.options.plugins being a function

barraponto:
We currently support postcss options being a function, but not options.plugins.
This should be mostly harmless -- I wanted to provide a test but we don't currently test support for any postcss plugin, do we?

### DIFF
--- a/lib/style-compiler/load-postcss-config.js
+++ b/lib/style-compiler/load-postcss-config.js
@@ -27,7 +27,9 @@ module.exports = function loadPostcssConfig (loaderContext, inlineConfig) {
     if (Array.isArray(inlineConfig)) {
       plugins = inlineConfig
     } else if (isObject(inlineConfig)) {
-      plugins = inlineConfig.plugins || []
+      plugins = (typeof inlineConfig.plugins === 'function')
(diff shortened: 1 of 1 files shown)


barraponto:
btw, i'm trying to follow the same pattern as postcss-loader: https://github.com/postcss/postcss-loader#plugins

gucong3000:
```
const loadOptions = require('postcss-load-options/lib/options.js');
const loadPlugins = require('postcss-load-plugins/lib/plugins.js');

plugins = loadPlugins(inlineConfig);
options = loadOptions(inlineConfig),
```
i think it is better to keep callback API sync with `postcss-load-config`

barraponto:
Well, it's not really my use case, but I can relate to a plugin configuration like this:

```javascript
{
  loader: 'postcss-loader',
  options: {
    plugins: (loader) => [
      require('postcss-import')({ root: loader.resourcePath }),
      require('postcss-cssnext')(),
      require('autoprefixer')(),
      require('cssnano')()
    ]
  }
}
```
Copied from https://github.com/postcss/postcss-loader#plugins

PR #868, opened 2017-07-01
Use `vue-loader` explictly when creating injector

anteriovieira:
Ref: https://github.com/vuejs/vue-loader/pull/837/files


kazupon:
Thanks!

PR #874, opened 2017-07-06
Fix template hmr, fix #863

### LINKED ISSUE #863: HMR stoped working on upgrade from 12.2.1 to 13
### Version
13.0.0

### Reproduction link
[https://github.com/select/audius](https://github.com/select/audius)

### Steps to reproduce
```
git clone git@github.com:select/audius.git
cd audius
npm i
npm start
```

HMR works for components e.g. `src/components/about-player.vue`

```
npm i -D vue-loader@13.0.0
```
HMR stops workging

### What is expected?
HMR updates work

HRM updates are shown in console but are not applied

---
As mentioned above the curious thing is that I can see the HMR update arrive in the browser but does not seem to update the page.

egoist:
seems we can't use `module.exports` while using `esModule` format, not sure if this is the correct fix though.

### DIFF
--- a/lib/template-compiler/index.js
+++ b/lib/template-compiler/index.js
@@ -75,7 +75,7 @@ module.exports = function (html) {
       '\nif (module.hot) {\n' +
       '  module.hot.accept()\n' +
       '  if (module.hot.data) {\n' +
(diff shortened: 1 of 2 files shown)


egoist:
Should be fine to merge it?

fnlctrl:
Probably should @yyx990803

PR #897, opened 2017-07-21
support comment option in template

Kingwl:
add a keyword ```comment``` to support  allow keep comments in sfc:  
```
<template comment>
  <div>
    ...
   <!-- comment -->
    ...
  </div>
</template>
```

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -89,10 +89,12 @@ module.exports = function (content) {
   var output = ''
   var parts = parse(content, fileName, this.sourceMap)
   var hasScoped = parts.styles.some(function (s) { return s.scoped })
+  var hasComment = parts.template && parts.template.attrs.comment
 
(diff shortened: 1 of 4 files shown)


yyx990803:
Changed the attribute to `comments` to keep it consistent with the `comments` option in core.

camslice:
Nice. This hasn't been added to the documentation yet as far as I can tell? Need assistance?

mariomka:
I would be useful to add it to the documentation.

JasperTey:
Is this feature available on vue-loader today? (I'm using 16.1.2)

I've been experiencing issues with HTML comments being stripped out from single file components, but only in the production build. I'm still not sure if it's a vue-loader specific behaviour, or something else in my webpack config, but this thread seems to indicate that I can control this by placing a "comments" attribute in the template tag as so: `<template comments>` to preserve the HTML comments within? That did not seem to make a difference, which confuses me.

PR #905, opened 2017-07-25
docs(options.md): Change default value of `esModule`

imcvampire:
Change default value from `false` to `true` follow new behavior in `>13.0.0`

### DIFF
--- a/docs/en/options.md
+++ b/docs/en/options.md
@@ -124,7 +124,7 @@ module.exports = {
 ### esModule
 
 - type: `boolean`
-- default: `false`
+- default: `tr`
 
   Whether to emit esModule compatible code. By default vue-loader will emit default export in commonjs format like `module.exports = ....`. When `esModule` is set to true, default export will be transpiled into `exports.__esModule = true; exports = ...`. Useful for interoperating with transpiler other than Babel, like TypeScript.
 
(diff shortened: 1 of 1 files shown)


docs/en/options.md
```diff
@@ -124,7 +124,7 @@ module.exports = {
 ### esModule
 
 - type: `boolean`
-- default: `false`
+- default: `tr`
```
francois2metz: You should say `true`.

imcvampire: Sorry. I fixed!

docs/en/options.md
```diff
@@ -124,7 +124,7 @@ module.exports = {
 ### esModule
 
 - type: `boolean`
-- default: `false`
+- default: `true`
```
kazupon: `vue-loader` is released v12.x later.
We should be mentioned about the version like the below:

```
default: `false` (v13.0+, `true`)
```

docs/en/options.md
```diff
@@ -116,6 +116,7 @@ module.exports = {
 
 - type: `boolean`
 - default: `true`
+- In v12.x or below, default value is `false`
```
kazupon: why?
please update the `esModule`.

imcvampire: I'm sorry :(. My mistake. I reverted this commit!

docs/en/options.md
imcvampire: I added the note about behaviour in v12.x or below.

PR #911, opened 2017-07-28
Add @ support for templates and styles

Atinux:
Hi!

I added `@` support for `<template>`  and `<style>`  since only `~` was working and `@` is more and more present in Vue.js projects.

Thank you :)


Atinux:
Ok after looking at the code, I wanted to add webpack alias support but you can do it by prefixing `~` in your files.

Let's say you have `@` alias, this will work: ``

PR #927, opened 2017-08-02
fix(transform-require): support ~/ aliases

pi0:
Hi. Recently we encounter a problem with Nuxt usage of aliases inside `.vue` files. The problem is that if aliases begin with `~/assets/logo.png` they will work with normal webpack requires but `vue-loader` tries to convert them to something like `/assets/logo.png` which is not a valid path. This PR fixes this condition and also adds appreciate tests.

Related issues:
- nuxt/nuxt.js#1241
- nuxt/nuxt.js#1257
- [this](https://github.com/nuxt/nuxt.js/commit/1f85b2dd13622575dbf0d2569d7486152d261d13) discussion

### DIFF
--- a/lib/template-compiler/modules/transform-require.js
+++ b/lib/template-compiler/modules/transform-require.js
@@ -40,7 +40,8 @@ function rewrite (attr, name) {
     var firstChar = value.charAt(1)
     if (firstChar === '.' || firstChar === '~') {
       if (firstChar === '~') {
-        value = '"' + value.slice(2)
+        var secondChar = value.charAt(2)
+        value = '"' + value.slice(secondChar === '/' ? 3 : 2)
       }
       attr.value = `require(${value})`
     }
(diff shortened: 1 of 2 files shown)


yuchonghua:
To support the point
:+1::+1::+1::+1::+1::+1:

Atinux:
Awesome :)

When do you plan to publish a patch for it?

/ping @yyx990803 @kazupon

HadiChen:
support

linyueb:
is so good,support!

liuyuanxiang:
support

yuchonghua:
When do you plan to publish a patch for it?

/ping @yyx990803 @kazupon

pi0:
`vue-loader@13.0.3` is released. Thanks @yyx990803 💟

yuchonghua:
Thanks @yyx990803 @kazupon 💟 And all your support :)

sodatea:
Correct me if I'm wrong, but seems `css-loader` no longer treats `~assets` & `~/assets` equally? Should `vue-loader` continue supporting this special case in the compiler?

PR #932, opened 2017-08-04
remove importLoaders from DEFAULT_OPTIONS

vanhoavn:
`importLoaders` is integer and need to be configurable by user.

https://github.com/webpack-contrib/css-loader/blob/master/lib/getImportPrefix.js


vanhoavn:
hi, any problem merging this PR?

PR #961, opened 2017-08-31
docs(EN,JA,RU): fix webpack template utils filename

lucaperret:
Replace the file path `build/util.js` by the correct `build/utils.js` of the [webpack template](https://github.com/vuejs-templates/webpack/blob/master/template/build/utils.js)


Jinjiang:
+1
Btw it seems `zh-cn` miss this section. I will create another PR for it.
Thanks.

PR #982, opened 2017-09-18
fix js-beautify version to 1.6.14, since js-beautify@1.7.0 is fucked

zyy7259:
vue-loader specify the version of js-beautify as ^1.6.14, which will allow the installation of js-beautify@1.7.0.

But package js-beautify@1.7.0 is not published correctly: https://github.com/beautify-web/js-beautify/issues/1247.

Temporarily fix its version to 1.6.14 can solve this problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ test/output
 docs/_book
 .DS_Store
 .idea
+.history
 *.iml
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "dependencies": {
     "consolidate": "^0.14.0",
     "hash-sum": "^1.0.2",
-    "js-beautify": "^1.6.14",
+    "js-beautify": "1.6.14",
     "loader-utils": "^1.1.0",
     "lru-cache": "^4.1.1",
     "postcss": "^6.0.6",


lmj0011:
this should solve issue I was having here:

https://github.com/beautify-web/js-beautify/issues/1247#issuecomment-330159973

LinusBorg:
I'm npot sure we should release a patch that freezes the version like that. people installing this version will be stuck with it.

This might be a solution if `js-beautify` doesn't resolve the problem in a timely manner, but we should wait out a little because the fix they need is very easy, the npm package is missing a few files they have to add again.

.gitignore
```diff
@@ -4,4 +4,5 @@ test/output
 docs/_book
 .DS_Store
 .idea
+.history
```
LinusBorg: unrelated change, should be removed.

package.json
```diff
@@ -43,7 +43,7 @@
   "dependencies": {
     "consolidate": "^0.14.0",
     "hash-sum": "^1.0.2",
-    "js-beautify": "^1.6.14",
+    "js-beautify": "1.6.14",
```
LinusBorg: we should do this once we see that `js-beautify` fails to actually fix it. We should wait a while.

Oliboy50:
nice try but => https://github.com/vuejs/vue-loader/commit/97bd432bba4350988bfd0db6d7a1a514b66a9826

bitwiseman:
@LinusBorg - The install is fixed. It is up to you if you want to switch to prettier permanently.

package.json
bitwiseman: Already fixed.  1.7.3 works.  The `js-beautify` project does use semantic versioning so `~1.7.3` would also be reasonable.

LinusBorg:
I will close this for now, will talk with the others about this.

PR #971, opened 2017-09-11
Support load postcss config by "config.path"

smadey:
ref #944

Issue #944, opened 2017-08-18
Ability to set postcss config path


justinhelmer:
### What problem does this feature solve?
lots of configs in a project (app config, webpack.base, webpack.client, webpack.server, vue-loader.config, css-loader.config, etc...) means it makes sense to have a `config/` folder in the project. It's unfortunate that `postcss.config.js` has to remain in the root, because there is no ability to specify the path.

The [postcss-loader options](https://github.com/postcss/postcss-loader#options) show a `config` option, that when used, allows the loader to look in a different location for `postcss.config.js`. 

A similar functionality would be very useful for `vue-loader`.

### What does the proposed API look like?
Another option to `vue-loader`:

```js
module.exports = {
  test: /\.vue$/,
  loader: 'vue-loader',
  options: {
    // ...
    postcss: {
      config: path.resolve(__dirname, './postcss.config.js')
    }
  }
};
```

LinusBorg:
instead of providing a path, simply require the config file and pass it to vue-loader's `postcss` option. That works already.

justinhelmer:
Thanks a ton for finagling this; I'll use the above approach until it's baked into the loader.

yyx990803:
Close via 94d374b318be33c91306631d0f1482f61a00e63d

PR #998, opened 2017-10-07
Support load postcss config by "config.path"

japboy:
ref #944

#971 could be right approach though it doesn't work with my case, so i made a different approach. idk what the right way is though. any helps would be appreciated.


yyx990803:
See https://github.com/vuejs/vue-loader/pull/971#issuecomment-335673917

PR #999, opened 2017-10-12
Rename functioanl-template.md to functional-template.md

Alex-Sokolov:
But not found where is links to this page


kazupon:
Good catch! Thanks!

Issue #1016, opened 2017-10-20
module IDs should be a digest of the contents


inopinatus:
### What problem does this feature solve?
Module IDs are currently generated by hashing part of the filename. This leads to inconsistency, since it assumes paths are the same on all servers, which they are not. We've had trouble with content delivery due to inconsistent manifests.

Systems like git and sprockets use sha for fingerprinting. Perhaps module IDs should be the same.

### What does the proposed API look like?
I am using this in production with my fork of vue-loader/lib/utils/gen-id.js:

```
// utility for generating a uid for each component file
// used in scoped CSS rewriting
const cache = Object.create(null)
const crypto = require('crypto')
const fs = require('fs')

function createHash (file, algorithm, encoding, maxlen) {
  var hash = crypto.createHash(algorithm)
  var contents = fs.readFileSync(file)
  hash.update(contents)
  return hash.digest(encoding).toLowerCase().trim().substring(0, maxlen)
}

module.exports = function genId (file) {
  return cache[file] || (cache[file] = createHash(file, 'sha256', 'hex', 10))
}
```

yyx990803:
Yes, this is something I realized a while ago but haven't implemented yet. If you'd like please make a PR!

yyx990803:
See https://github.com/vuejs/vue-loader/pull/1018#issuecomment-341797205

PR #1018, opened 2017-10-21
Use sha digest of file to generate moduleId.

inopinatus:
implements feature suggestion in #1016.


yyx990803:
Thanks for the PR! Unfortunately content-based hashing can only be used in production mode, otherwise it breaks hot-reloading during development. It is now enabled for production builds in 13.4.0.

PR #1030, opened 2017-10-31
fix: scoping of multiple animations

Damell:
fixes https://github.com/vuejs/vue-loader/issues/1031


Damell:
fixes https://github.com/vuejs/vue-loader/issues/1031

PR #1021, opened 2017-10-22
Fix Error which is braking absolute paths (Issue: #1005)

progressman:
Fix a typo in line 50 that leads to an incorrect path submitted to require.
Return unchanged url for absolute paths.

PR #1038, opened 2017-11-03
fix srcset transform module not resovling alias path

lemures-t:
srcset transform module cannot handle alias path correctly.
e.g.
```javascript
<img srcset="~assets/logo@2x.png 2x" />
```
It is caused by parsing url started with ```~``` in ```/lib/template-compiler/modules/transform-srcset.js```, line 50.
Fixed with more test unit is added in ```test.js```
After all, many thanks to @AndreKR 's adding support for ```srcset```


lemures-t:
Fixed by #1021 and already released by v13.4.0. Closed.

Issue #942, opened 2017-08-11
Warn when using with postcss-loader


coolzjy:
### Version
12.2.0

### Reproduction link
[https://github.com/coolzjy/vue-loader-postcss](https://github.com/coolzjy/vue-loader-postcss)

### Steps to reproduce
1. Init a project using vue-cli and webpack template
2. Install `postcss` and `postcss-loader`
3. Edit `build/utils.js`:
```js
-    css: generateLoaders(),
+    css: generateLoaders('postcss'),
```
4. `npm run dev`

### What is expected?
Works fine

### What is actually happening?
Got warn
![image](https://user-images.githubusercontent.com/3762377/29204836-cd5d28f8-7eab-11e7-8c55-95d252c9e8c1.png)

coolzjy:
have tried:
1. set `sourceMap: false` in postcss-loader config
2. set `cssSourceMap: false` in vue-loader config
Still not resolved.

Related issues:
https://github.com/postcss/postcss-loader/issues/253
https://github.com/postcss/postcss-loader/issues/264

applefreak:
Has this ever got solved?

StarXY:
Hi,
I had the same issue with you.
my repo: [https://github.com/StarXY/webpackDemo](url)
version: 13.3.0
nodejs: 8.2
OS: win7

StarXY:
maybe a temporary solution. 
from
```
<style lang="postcss"></style>
```
to
```
<style ></style>
```

related: https://github.com/vuejs/vue-loader/pull/937

yyx990803:
There's no need to use `postcss-loader` with `vue-loader`. Just use `vue-loader`' own postcss option.

blake-newman:
@yyx990803 If not using `lang="postcss"` ect most IDE layers do not read the contents as postcss but plain css. Thus syntax highlighting is broken. This may be why people are configuring there own loaders which causes warning. I have created a Fix with #1055

Issue #800, opened 2017-04-30
Disable postcss config file when using inline config


egoist:
### What problem does this feature solve?
I manually load postcss config from its config file and set `postcss` in vue loader options, but `vue-loader` will also load it and merge inline config with config file so it leads to duplicated plugins.

btw, `postcss-loader` does not use config file when inline config is set.

### What does the proposed API look like?
Maybe this should be the default behavior, or we can introduce an option for this.

barraponto:
Agreed. It would be very useful for `neutrino-middleware-vue`.

barraponto:
btw, can we close in favor of #654 ?

Issue #654, opened 2017-02-18
Disable auto-loading PostCSS config file


egoist:
Since `postcss-loader` does not load it when option is set. https://github.com/postcss/postcss-loader/blob/master/index.js#L67

michael-ciniawsky:
@egoist I work on a PR already including a few other smaller improvements and yes I would also change it to that logic. `inline ? use(inline) : use('postcss.config.js')  // inline = webpack.config.js`. So it's either explicit via `webpack.config.js` or implicit via `postcss.config.js, but clearly separated.

PR #1055, opened 2017-11-12
Feature: Handle postcss file extensions if no loader options give

blake-newman:
~~= Problem~~

~~- vue-loader automatically handles postcss by default, there is many issues
around using custom postcss lang option.~~
~~- Using postcss without a `lang` attribute causes SFC to loose syntax
highlghting~~

~~= Solution~~

~~- Enable option to alias langs to loader settings.~~
~~- specify `postcss` in lang and maintain default loader configurations~~

## Edit
If no postcss loader is provided in vue loader options then default to
vue-loader handling

This allows you to use vue internal postcss/css handling with known
postcss extensions

fixes: #942
fixes: #800
fixes: #654

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -29,6 +29,10 @@ const defaultLang = {
   script: 'js'
 }
 
+const postcssExtensions = [
+  'postcss', 'pcss', 'sugarss', 'sss'
+]
+
 // When extracting parts from the source vue file, we want to apply the
 // loaders chained before vue-loader, but exclude some loaders that simply
 // produces side effects such as linting.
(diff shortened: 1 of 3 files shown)


octref:
I'm not against using postcss grammar to highlight css, so `<style>` by default gets postcss highlight.
However the issue is `vscode-css-languageservice` does not handle postcss. So autocomplete and other language features won't work.

> Vue automatically handles postcss by default

I believe that's a part of Vue's specific webpack configuration in the webpack template. I don't think Vue handles postcss by default (correct me if I'm wrong).

This has been suggested before: https://github.com/vuejs/vue-loader/pull/500
My opinion is still the same -- being explicit on `lang` is better than implicit.
And from Vetur side, it's much easier to build language services according to a spec instead of according to user's ever-changing webpack config.

blake-newman:
@octref 

Yes sorry `vue-loader` handles postcss by default, using `lang="postcss"` has known issues as far as i'm aware and i have also not been able to configure it correctly (perhaps doing something wrong).

I was also against #500, this solution does not remove explicitness from the attribute. You still have to provide the lang attribute. It merely aliases a lang to another. `<style lang="postcss">` with the alias set to `{ postcss: 'css' }` will run lang postcss with vue internals.

I do believe if possible moving into the Vetur will be best approach, as far as i could work out it was not a trivial thing to implement. 

On further thoughts, we should internally handle aliasing `postcss` to use vue-loaders internals unless overridden in `loader` configuration.

blake-newman:
I don't think adding a public API is good, as it could lead to bad implementations now thinking about it.

PR #1056, opened 2017-11-13
[DO NOT MERGE] POC using vue-component-compiler

znck:
Related PR https://github.com/vuejs/vue-component-compiler/pull/29


yyx990803:
We'll keep v14 as-is to avoid possible regressions. For v15 let's keep working with https://github.com/vuejs/component-compiler-utils

PR #1070, opened 2017-11-20
only generate css source map when cssSourceMap is enabled

melonHuang:
Please only generate css source map when cssSourceMap is enabled. 

Or it will cause the postcss-loader to throw a warning: `Previous source map found, but options.sourceMap isn't set`

The warning is triggered under this condition
```
if (sourceMap == undefined && map) {
      this.emitWarning(`\n\n ⚠️  PostCSS Loader\n\nPrevious source map found, but options.sourceMap isn't set.\nIn this case the loader will discard the source map entirely for performance reasons.\nSee https://github.com/postcss/postcss-loader#sourcemap for more information.\n\n`)
}
```

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -97,7 +97,7 @@ module.exports = function (content) {
 
   let output = ''
   const bustCache = !isProduction && options.cacheBusting !== false
-  const parts = parse(content, fileName, this.sourceMap, sourceRoot, bustCache)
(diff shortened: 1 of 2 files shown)


yyx990803:
The check had to be a bit more consistent with the CSS source map conditions, so I landed this with some tweaks. But thanks for making the PR!

PR #1081, opened 2017-12-02
[Fixed #1064] supported ES default exported function for custom block in runtime

### LINKED ISSUE #1064: custom block with es6 content
### Version
13.5.0

### Reproduction link
[https://github.com/dunkfordyce/custom-block-es6](https://github.com/dunkfordyce/custom-block-es6)

### Steps to reproduce
Run the sample repo as usual. The custom block `foo` has a loader build/foo-loader.js. The code exports a default function to modify the Component. Because the code is ES6 the function is never run as the loader does not check for an ES6 module, it sees an object instead of a function. 

### What is expected?
The default exported function should be run.

The vue-loader code sees an object instead of a function.

Jinjiang:
#1064

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -367,8 +367,11 @@ module.exports = function (content) {
 
         output +=
           'var customBlock = ' + requireString + '\n' +
-          'if (typeof customBlock === "function") {' +
-          '  customBlock(Component)' +
+          'if (customBlock && customBlock.__esModule) {\n' +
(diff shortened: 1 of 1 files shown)


yyx990803:
Can we add a test case for this?

Jinjiang:
Yes, it should have test cases. I will add them soon.
Thanks.

yyx990803:
I added them :)

PR #1082, opened 2017-12-04
lang=sass not supported

llKreator:
changed sass to scss


yyx990803:
There's probably something wrong with your configuration but `lang="sass"` definitely should work. If that's a bug we will need a proper bug report with reproduction. Thanks.

PR #1088, opened 2017-12-09
[docs] updated status of inject-loader (close #892)

### LINKED ISSUE #892: Inject-Loader 3.0.0 warning is out of date
### Version
12.2.2

### Reproduction link
[https://vue-loader.vuejs.org/en/workflow/testing-with-mocks.html](https://vue-loader.vuejs.org/en/workflow/testing-with-mocks.html)

### Steps to reproduce
The documentation states that Inject-Loader 3.0.0 is unstable (circa September 2016), but 3.0.0 has been out of beta since March 2017, so it seems like this warning is out of date.

### What is expected?
No warning about inject-loader 3.0.0

Warning about inject-loader 3.0.0

Jinjiang:
1. Removed the unstable note about `inject-loader@3.x`.
2. Note for disabling `esModule` option in inject mode because `import/export` must be top-level.

Ref: #892 

Thanks.

### DIFF
--- a/docs/en/workflow/testing-with-mocks.md
+++ b/docs/en/workflow/testing-with-mocks.md
@@ -4,6 +4,8 @@ In a real world application, our components most likely have external dependenci
 
 `vue-loader` provides a feature that allows you to inject arbitrary dependencies to a `*.vue` component, using [inject-loader](https://github.com/plasticine/inject-loader). The general idea is that instead of directly importing the component module, we use `inject-loader` to create a "module factory" function for that module. When this function gets called with an object of mocks, it returns an instance of the module with the mocks injected.
 
+> Note: You must disable `esModule` option in inject mode, or you will get an error.
+
 Suppose we have a component like this:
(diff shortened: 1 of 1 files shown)


docs/en/workflow/testing-with-mocks.md
```diff
@@ -14,9 +16,9 @@ Suppose we have a component like this:
 
 <script>
 // this dependency needs to be mocked
-import SomeService from '../service'
+const SomeService = require('../service')
 
-export default {
+module.exports = {
   data () {
```
kazupon: In ES5, I think you should be change to `data: function () {`.  :)

Jinjiang: Actually it just disables `esModule` option, not the whole ES6+ syntax. So that's OK imo :-)

kazupon: okay! :)

PR #1095, opened 2017-12-18
[fixed #965] more friendly error message for loading PostCSS config

### LINKED ISSUE #965: Completely ignoring .postcssrc.js
### Version
13.0.4

### Reproduction link
[https://github.com/warent/vue-loader-bug1](https://github.com/warent/vue-loader-bug1)

### Steps to reproduce
1. npm install
2. npm run build
3. npm start

### What is expected?
Should throw an error because .postcssrc.js has bogus data

Everything runs fine

Jinjiang:
Ref: #965 
Thanks.


Jinjiang:
Additional changes:

* Removed duplicated error reports in `bundle()`.
* Supported one more argument `wontThrowError` in `test()` and `bundle()` to make it possible to analyze error/warning reports.
* Differentiated two test cases with same `"postcss options"` description before.

Thanks.

PR #1101, opened 2017-12-22
Sorted all current test cases into some groups

Jinjiang:
... to make tests more maintainable.

The structure after sorting:

* vue-loader
  * basic usage
    * basic
    * pre-processors
    * style import
    * template import
    * script import
  * advanced features
    * pre/post loaders
    * support chaining with other loaders
    * expose filename
    * extract CSS
    * extract CSS using option
    * extract CSS using option (passing plugin instance)
    * pre-processors with extract css
    * source map
    * dependency injection
    * hot reload enabled
    * hot reload disabled
    * cacheBusting: false
  * template block features
    * template with comments
    * transpile ES2015 features in template
    * translates relative URLs and respects resolve alias
    * transformToRequire option
    * should allow adding custom html loaders
    * custom compiler modules
    * custom compiler modules (string)
    * custom compiler directives
    * functional component with styles
    * functional template
  * style block features
    * scoped style
    * media-query
    * supports-query
    * postcss options
    * load postcss config file
    * load postcss config file by path
    * load postcss config file with js syntax error
    * postcss lang
    * css-modules
  * script block features
    * allows to export extended constructor
    * support es compatible modules
    * named exports
  * custom block features
    * extract custom blocks to a separate file
    * add custom blocks to the webpack output
    * custom blocks work with src imports
    * passes Component to custom block loaders
    * custom blocks can be ignored
    * custom blocks with ES module default export
    * passes attributes as options to the loader
    * pre/post loaders for custom blocks
  * SSR
    * css-modules in SSR
    * SSR style and moduleId extraction

Thanks.


yyx990803:
Looks like there's a conflict - I think it would be even better to split the tests and helpers into separate files - right now a single file is becoming hard to navigate.

Jinjiang:
Agree, I will resolve the conflict and separate the code soon.

PR #1112, opened 2018-01-05
add babel configure options

joebnb:
```
<template>
    <div>
        <test1 />
        <test2 />
    </div>
</template>
<script>
    const test1 = ()=>import('./component/a.vue');
    const test2 = ()=>import('./component/b.vue');
    export default {
        components:{
            test1,test2
        }
    }
</script>
```
fix some case if user need configure  bable option,like upper code sentence,i need configure  `syntax-dynamic-import` to reslove dynamic import or other problem.

code refferred from loader.js bubleOptions.

and then config like this:
```
{
                test: /\.vue$/,
                exclude: /(node_modules|bower_components)/,
                use: {
                    loader: 'vue-loader',
                    options:{
                        bable:{
                            presets: ['babel-preset-env'],
                            plugins:[
                                'syntax-dynamic-import'
                            ]
                        }
                    }
                }
            }
```

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -106,8 +106,7 @@ module.exports = function (content) {
   const babelOptions =
     hasBabel && options.bable ? '?' + JSON.stringify(options.bable) : ''
 
-
-    let output = ''
+  let output = ''
 
   const parts = parse(
     content,
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -103,7 +103,11 @@ module.exports = function (content) {
   const bubleOptions =
     hasBuble && options.buble ? '?' + JSON.stringify(options.buble) : ''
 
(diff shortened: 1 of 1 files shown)


Jinjiang:
Is that OK to set the config directly in `.babelrc` file? Because this config only affects `*.vue` files but doesn't affect `*.js` files. This kind of config is general enough to put it out imo.

Btw. The buble config in `loader.js` is only for expression in `<template>` in `*.vue` files. So it's just good to configure in `vue-loader` options.

Thanks.

yyx990803:
We expose Buble options because Buble doesn't have an equivalent of `.babelrc`. For configuring Babel, always use `.babelrc` for consistent config between Vue and normal JavaScript files.

PR #1123, opened 2018-01-12
move vue-template-compiler module dependency from devDependencies to dependencies

nicocevallos:
I made this PR in order to avoid to add vue-template-compiler module to the package.json in the projects that use the vue-loader due they don't have a direct relationship to vue-template-compiler


nicocevallos:
Oh, ok, thanks for the info, I didn't know it.
Closing this PR.

PR #1126, opened 2018-01-15
Update link and pic

totalamd:


yyx990803:
Thanks for the PR! We will do an update for all docs using assets suggested by webpack team.

PR #1128, opened 2018-01-15
orpho

totalamd:


yyx990803:
Looks like the change is already in the codebase... not sure why, but thanks for the PR anyway.

PR #1131, opened 2018-01-18
Support multiple rule definitions

HaoyCn:
In my case I need to define different loader options, one for my own codes, and the other one for some third-party codes. Howerver, vue-loader repeatedly mounts `__vueOptions__` to `this.options` and finally options get overridden. Now I pass an  `optionId` to internal compilers to resolve the correct one.

An example is given in the test.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -69,16 +69,21 @@ module.exports = function (content) {
     options.esModule = false
   }
 
-  let vueOptionId = 0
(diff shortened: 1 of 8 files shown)


yyx990803:
Unfortunately I don't think the strategy of defining custom properties on `this.options` works anymore in webpack 4. `vue-loader` 14 now defines it on `this._compilation`, but that also causes problems when used with `thread-loader` or Happypack, so we are still investigating what's the best future-proof way to deal with this. Thanks for the PR anyway.

PR #1143, opened 2018-01-31
always return undefined when calling callback()

AielloChan:
Refer to [webpack synchronous-loaders](https://webpack.js.org/api/loaders/#synchronous-loaders)


AielloChan:
You are right, but it just a official standard, like script indent.

Jinjiang:
I would say you may misunderstand the “standanard”. That’s either return **or** this.callback, not both of them.

yyx990803:
Thanks, I don't think this is not necessary since the function implicitly returns `undefined` anyway. I think webpack docs probably meant you need to make sure not to accidentally return a non-`undefined` value when using the callback.

Issue #624, opened 2017-02-06
8.6.1 : 'TypeError: loader.charAt is not a function' in webpack 2.2.1


DeepenLau:
I use webpack2.2.1 in vue1.0.28 , when i use extract-text-webpack-plugin in webpack configuration like this 
```js
  ...
  var ExtractTextPlugin = require('extract-text-webpack-plugin')

  module: {
    rules: [
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          loaders: {
            stylus: ExtractTextPlugin.extract({
              use: [
                'css-loader?{discardComments:{removeAll:true}}',
                'stylus-loader'
              ],
              fallback: 'vue-style-loader'
            })
          },
          postcss: [
            require('autoprefixer')({
              browsers: ['> 0.1% in CN']
            })
          ]
        }
      }
    ]
  },
  ...
  plugins: [
    ...
    new ExtractTextPlugin({
      filename: '[name].[contenthash:8].css',
      disable: false,
      allChunks: true
    })
    ...
  ]
  ...

```
and it show me  this message:
**"Module build failed: TypeError: loader.charAt is not a function"**
when i remove the  extract-text-webpack-plugin like this
```js
  module: {
    rules: [
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          loaders: {
            stylus: 'vue-style-loader!css-loader!stylus-loader'
          },
          postcss: [
            require('autoprefixer')({
              browsers: ['> 0.1% in CN']
            })
          ]
        }
      }
    ]
  }
```
then it works.
by the way, webpack2.2.1 must use the new version of extract-text-webpack-plugin, so the latest version is
 "extract-text-webpack-plugin": "^2.0.0-rc.3",
so, i do not know it is my wrong in somewhere or it is a bug.
please help, thanks a lot!

---
update at 19:37 in china
I switched the version of "extract-text-webpack-plugin" to "^2.0.0-beta.5" and change my configuration in this code
```js
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          loaders: {
            stylus: ExtractTextPlugin.extract({
              loader: 'css-loader?{discardComments:{removeAll:true}}!stylus-loader',
              fallback: 'vue-style-loader'
            })
          },
          postcss: [
            require('autoprefixer')({
              browsers: ['> 0.1% in CN']
            })
          ]
        }
      }
```
finally, "extract-text-webpack-plugin" works

[...]

PR #1166, opened 2018-02-19
Fix for 'loader.charAt is not a function' in v8

natocTo:
I have some older project still runnig on Vue 1 and crash on 'loader.charAt is not a function' error like in the #624 issue.
I am not javascript expert, but this edit solve my issue and maybe can solve also for other.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -171,6 +171,11 @@ module.exports = function (content) {
   }
 
   function ensureBang (loader) {
+    if (Array.isArray(loader)) {
+      loader = stringifyLoaders(loader)
+    } else if (typeof loader === 'object') {
+      loader = stringifyLoaders([loader])
+    }
     if (loader.charAt(loader.length - 1) !== '!') {
       return loader + '!'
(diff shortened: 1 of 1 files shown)


natocTo:
Thanks for reply. I am not quite similiar with testing but I can look at it..

Btw. same logic I edded here (8.x is last loader for Vue 1) you can found in version 12 here https://github.com/vuejs/vue-loader/blob/v12/lib/loader.js#L484 . The function is same. Maybe can you found test there. :)

natocTo:
So I am not able to test this, I am sorry.

I clone this repo. Switch the branch to 8x, run npm install and npm test. Even without this issue I end up with 6+ failing tests. So I do not know how fix all this tests and even write this new one.

If this can not be able to merged without test, maybe we can close it. But same logic is used in newer version and for me this fix the issue 'loader.charAt is not a function'.

Jinjiang:
OK, that's alright. I will have a deep look at it. :-)
Thanks.

PR #1174, opened 2018-02-26
Update asset-url.md

tomwang1013:
It only work without the `/` after `~` in my project with `vue-loader@14.1.1`. Is this a typo or something error?
P.S. i use `webstorm@2017.3.2`. But it shouldn't matter.


yyx990803:
That's a typo indeed...

Issue #1171, opened 2018-02-21
vue-loader 14.1.0 ignores transformToRequire config


andreyrd:
### Version
14.1.0

### Reproduction link
[https://github.com/andreyrd/vue-loader-transformtorequire-demo](https://github.com/andreyrd/vue-loader-transformtorequire-demo)

### Steps to reproduce
1. Clone the repo.
2. Run `yarn install` to install the project with `vue-loader: 14.0.0`.
3. Run `yarn dev` and note that there are two Vue logos showing up, one after the other.
4. Stop the server.
5. Edit the `package.json` and update `vue-loader` to `14.1.0`.
6. Run `yarn install` to install `vue-loader: 14.1.0`.
7. Run `yarn dev` and note that the second Vue logo no longer shows up.

### What is expected?
The second image should be showing up.
The second image is a custom component that relies on the transformToRequire config to have `Test: 'source'` in order to transform it's `source` prop into a `require()` statement.

See `build/vue-loader.conf.js` for the config I'm talking about.

### What is actually happening?
Since 14.1.0, `vue-loader` is ignoring the `transformToRequire` property in the config.

---
I believe the bug was introduced here: https://github.com/vuejs/vue-loader/commit/289b4d9e7c4d2761597c9f038332897abb5c3f76#diff-7c7d4918b118d27d18538f53b8155e88L138

oliverzy:
yes, today I also found this issue after upgrading vue-loader.
It can be fixed by changing https://github.com/vuejs/vue-loader/blob/master/lib/template-compiler/index.js#L17 to
```
const defaultModules = [transformRequire(vueOptions.transformToRequire), transformSrcset()]
```

yyx990803:
Fixed via #1175

PR #1175, opened 2018-02-26
fix: get transformToRequire from vueOptions instead of options for template-compiler (#1171)

fnlctrl:
`transformToRequire` is actually broken since v14 because compiler options is passed by `this._compilation.__vueOptions__`, instead of webpack loader option, so it should be `vueOptions.transformToRequire` here. 
Not sure why tests didn't fail for `options.transformToRequire` :/


oliverzy:
The related issue is here: #1171.
Is it possible to push a fix for this one?

fnlctrl:
@oliverzy This PR should fix that issue already. 😄

PR #1192, opened 2018-03-12
Added link to component options spec

kraigh:
I'm learning Vue and had to look up what these options were, so I thought it might be helpful to direct others to the appropriate docs page.

### DIFF
--- a/docs/en/start/spec.md
+++ b/docs/en/start/spec.md
@@ -58,7 +58,7 @@ More details can be found in [Using Pre-Processors](../configurations/pre-proces
 
 - The script is executed in a CommonJS like environment (just like a normal `.js` module bundled via webpack), which means you can `require()` other dependencies. And with ES2015 support, you can also use the `import` and `export` syntax.
 
-- The script must export a Vue.js [component options object](https://012.vuejs.org/api/options.html). Exporting an extended constructor created by `Vue.extend()` is also supported, but a plain object is preferred.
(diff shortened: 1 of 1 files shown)


docs/en/start/spec.md
```diff
@@ -58,7 +58,7 @@ More details can be found in [Using Pre-Processors](../configurations/pre-proces
 
 - The script is executed in a CommonJS like environment (just like a normal `.js` module bundled via webpack), which means you can `require()` other dependencies. And with ES2015 support, you can also use the `import` and `export` syntax.
 
-- The script must export a Vue.js component options object. Exporting an extended constructor created by `Vue.extend()` is also supported, but a plain object is preferred.
+- The script must export a Vue.js [component options object](https://012.vuejs.org/api/options.html). Exporting an extended constructor created by `Vue.extend()` is also supported, but a plain object is preferred.
```
Jinjiang: The link you provide is just about the legacy version.
Maybe this link is better: https://vuejs.org/v2/api/#Options-Data
or we can add some more anchors like "Global" which contains "Global API" and "Global Config", "Options" which contains "Options / *" and Instance, Template ....
@chrisvfritz how do you think about it? :-)
Thanks.

chrisvfritz: I think we can I think we can link to the first section for Instance options as a starting place:

https://vuejs.org/v2/api/#Options-Data

PR #1200, opened 2018-03-21
Fix errors using loader with webpack 4.2

cseufert:


lib/style-rewriter.js
```diff
@@ -33,7 +33,7 @@ module.exports = function (css, map) {
   var cb = this.async()
 
   var query = loaderUtils.getOptions(this) || {}
-  var options = this.options.vue || {}
+  var options = this.options && this.options.vue || {}
```
davidpelayo: In direct assignments you may be have to double coerce the first logic comparison to ensure it returns a boolean and fallbacks to `{}` if resolves to false. So it would be:
```js
var options = !!this.options && this.optionsvue || {}
```

PR #1203, opened 2018-03-23
Remove `include` before normalizing loaders, fixes #1201

### LINKED ISSUE #1201: v15-beta fails when the vue-loader rule uses `include`
### Version
15.0.0-beta.2

### Reproduction link
[https://github.com/Loilo/vue-loader-15-include-repro](https://github.com/Loilo/vue-loader-15-include-repro)

### Steps to reproduce
**Clone the repository and install npm dependencies:**
```
git clone https://github.com/Loilo/vue-loader-15-include-repro.git
cd vue-loader-15-include-repro
npm install
```

**Run webpack:**
```
npx webpack
```

### What is expected?
The `VueLoaderPlugin` should initialize correctly.

The `VueLoaderPlugin` complains that "no matching rule for vue files are found".

---
I've dug into the source and found the cause: The check if a loader [is matching `foo.vue`](https://github.com/vuejs/vue-loader/blob/ab067b0/lib/plugin.js#L14) returns `false` if the rule has an `include` property. This is correct for webpack's own behaviour but probably undesirable for `vue-loader`.

A sufficient solution would probably be to (locally) remove the `include` property before normalizing rules. <s>I'd be happy to attach a PR for this.</s> I have attached a PR for this.

### LINKED ISSUE #1202: [15.0.0-beta.2] VueLoaderPlugin Error: no matching rule for vue files are found.
### Version
15.0.0-beta.2

### Reproduction link
n/a

### Steps to reproduce

I noticed that i am getting this error when rule for `.vue` files contains `include` option with absolute folder path. Example of failing webpack config

```js
const path = require('path');
const { VueLoaderPlugin } = require('vue-loader');

const src = path.resolve(__dirname, 'src');

module.exports = {
  entry: {
    app: path.join(src, 'index.js'),
  },
  output: {
    publicPath: '/',
    path: path.resolve(__dirname, 'build'),
    filename: '[name].bundle.js',
    chunkFilename: '[id].[name].chunk.js',
  },
  resolve: {
    extensions: [
      '.js',
      '.vue',
    ],
  },
  module: {
    rules: [
      {
        test: /\.js$/,
        include: src,
        loader: 'babel-loader',
      },
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        include: src, // <-- causes the error
        options: {
          preserveWhitespace: false,
        },
      },
    ],
  },
  plugins: [
    new VueLoaderPlugin(),
  ],
};

```

Full log:

```
Error: VueLoaderPlugin Error: no matching rule for vue files are found.
    at VueLoaderPlugin.apply (C:\dev\vue-loader-next\node_modules\vue-loader\lib\plugin.js:19:13)
    at Compiler.apply (C:\dev\vue-loader-next\node_modules\tapable\lib\Tapable.js:375:16)
    at webpack (C:\dev\vue-loader-next\node_modules\webpack\lib\webpack.js:33:19)
    at processOptions (C:\dev\vue-loader-next\node_modules\webpack\bin\webpack.js:335:15)
    at yargs.parse (C:\dev\vue-loader-next\node_modules\webpack\bin\webpack.js:397:2)
    at Object.Yargs.self.parse (C:\dev\vue-loader-next\node_modules\yargs\yargs.js:533:18)
    at Object.<anonymous> (C:\dev\vue-loader-next\node_modules\webpack\bin\webpack.js:152:7)
    at Module._compile (module.js:662:30)
    at Object.Module._extensions..js (module.js:673:10)
    at Module.load (module.js:575:32)
    at tryModuleLoad (module.js:515:12)
    at Function.Module._load (module.js:507:3)
    at Function.Module.runMain (module.js:703:10)
    at startup (bootstrap_node.js:193:16)
    at bootstrap_node.js:665:3
```

Webpack version **3.11.0**

### What is expected?
I would like to bundle my project without error.

Webpack throws mentioned error

Loilo:
This fixes #1201 by creating a local copy of rules which use the `include` option and then removes it property from that copy.

Rules without the `include` option are not copied to spare needless operations.

**EDIT:** Also fixes #1202 (which is a duplicate of #1201).

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -6,8 +6,18 @@ module.exports = class VueLoaderPlugin {
   apply (compiler) {
     // get a hold of the raw rules
     const rawRules = compiler.options.module.rules
+    // remove `include` property, see #1201
+    const rawRulesWithoutInclude = rawRules.map(rule => {
+      if ('include' in rule) {
+        const copy = Object.assign({}, rule)
+        delete copy.include
+        return copy
+      } else {
+        return rule
+      }
+    })
     // use webpack's RuleSet utility to normalize user rules
-    const rawNormalizedRules = new RuleSet(rawRules).rules
+    const rawNormalizedRules = new RuleSet(rawRulesWithoutInclude).rules
 
     // find the rule that applies to vue files
     const vueRuleIndex = rawRules.findIndex((rule, i) => {


Loilo:
In case anyone wants to check it out: I've added the [`with-fix-1203`](https://github.com/Loilo/vue-loader-15-include-repro/tree/with-fix-1203) branch to the #1201 reproduction repository, which uses this PR to fix the issue.

yyx990803:
Thanks for the PR! Unfortunately the fix isn't entirely safe because this would cause all cloned rules to lose the `include` check. I've implemented a fix with test in 2be5507

PR #1208, opened 2018-03-23
feat(loader): support options.productionMode

pi0:
With webpack 4, `this.minimize` value is `undefined`. This PR allows enabling `minimize` with loader options.

----

A side note from [here](https://medium.com/webpack/webpack-4-migration-guide-for-plugins-loaders-20a79b927202):
> Guideline: Loaders should receive all options via this.query. They should not use other ways to receive options, i. e. no property in webpack options, no environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,3 +222,4 @@ The following option has been changed to resourceQuery:
 - `transformAssetUrls`
 - `optimizeSSR`
 - `hotReload`
+- `productionMode`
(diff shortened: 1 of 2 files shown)


yyx990803:
I don't think the option should be called `minimize` since `vue-loader` itself doesn't perform any minification. `productionMode` is more explicit.

pi0:
@yyx990803 done :)

PR #1215, opened 2018-03-24
Notice Template Handle for v15

JounQin:
I don't know is the description accurate, but it works to me.


yyx990803:
Thanks for the catch - I decided to drop the consolidate usage and use webpack loaders for template preprocessing as well. See [updated notes](https://github.com/vuejs/vue-loader/tree/next#template-preprocessing)

PR #1218, opened 2018-03-27
add .d.ts declaration file

JounQin:
enable to work with `webpack.config.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "15.0.0-beta.7",
   "description": "Vue single-file component loader for Webpack",
   "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
   "scripts": {
     "test": "jest --env node",
     "lint": "eslint lib test --fix",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "15.0.0-beta.7",
   "description": "Vue single-file component loader for Webpack",
   "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
```
ktsn: Can we change it to `types` field since looks like `types` is the recommendation now?
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

JounQin: I used `types` before, but I find out `typings` in `vue` core [package.json](https://github.com/vuejs/vue/blob/dev/package.json#L9) and `vue-router`, should it be consistent here?

ktsn: Oh, I see. I guess it's ok to leave as `typings` then 🙂

PR #1289, opened 2018-05-07
Pull in upstream changes

davidrunger:

PR #1290, opened 2018-05-07
Pull in upstream changes

davidrunger:


davidrunger:
I am an idiot. This is my second unintentional PR. (See also #1289.) I am so sorry.

PR #1302, opened 2018-05-17
fix: check on "null"

LasaleFamine:
Just found this "ugly" checks while reading the code. I hope it's not wanted.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -126,10 +126,10 @@ function cloneRule (rule) {
     },
     resourceQuery: query => {
       const parsed = qs.parse(query.slice(1))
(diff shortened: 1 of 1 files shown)


Justineo:
`parsed.vue === null` here is a shorthand for `parsed.value !== null && parsed.value !== undefined` and we actually have [a configured ESLint rule](https://github.com/vuejs/eslint-plugin-vue-libs/blob/a3b63ec188d504e1421c9437833dd3f92fdb358a/config.js#L35) for this case (to allow the used of `==` against `null`). Thanks.

LasaleFamine:
I didn't quite understand this mechanism, but ok then! Sorry for bothering.

PR #1313, opened 2018-05-24
Inject issuerPath to resourceQuery for custom blocks

egoist:
When using `src` attribute on custom blocks, the requested file won't be able to know which file is actually requesting it, maybe we can directly inject `resourcePath` to the query as a workaround.


yyx990803:
Let only add this if it is a `src` import. Otherwise the request string will get very long and make error stacks hard to read.

PR #1318, opened 2018-05-25
Allow use of ':name' attribute.

ibanezm:
Avoids forcing Javascript language when using lang="ts" in Jetbrains products.


yyx990803:
I don't use Jetbrain products, can you explain what problem does this solve?

ibanezm:
Im sorry, it seems to be a bug in Jetbrains products itself. _Nothing_ about using "lang" attribute.
Sorry.

Dont merge, as it duplicates query params when using both attributes simultaneously and can be confusing (ie. "<script :lang="ts" lang="js"> builds something like "?lang=ts&lang=js").

pd. sorry for lossing your time, but thanks for replying :)

PR #1319, opened 2018-05-28
fix: default parser was removed from prettier

CasparChou:
same as [fix: default parser was removed from prettier](https://github.com/vuejs/component-compiler-utils/pull/15)


channg:
Where can I find the repaired version number?

azz:
We've actually pushed a patch version of Prettier that squelches this error, so if you upgrade either prettier `npm upgrade --depth=100 prettier` it should work. (`prettier@1.13.2`)

PR #1326, opened 2018-05-28
fix(v14): prettier default parser removal

pwang2:
same issue steamed down from https://github.com/prettier/prettier/issues/4567 for legacy `vue-loader` version


pwang2:
Close as already hotfixed by https://github.com/vuejs/vue-loader/pull/1323

PR #1329, opened 2018-05-29
Module build failed: Error: No parser and no file path given, couldn't infer a parser.

maplej:
Prettier has caused "Module build failed: Error: No parser and no file path given, couldn't infer a parser." in their 1.13.0 update. Downgrade to the previous version to fix this error.


ulivz:
Duplicate of #1319

PR #1334, opened 2018-06-01
fix unexpected error when options of cache-loader contains !

ulivz:
# Summary

When `cacheIdentifier` contains exclamation mark `!`, the `!` would be treated as the separator between different loaders. which will cause unexpected compilation error.

e.g.

when you have these options set:

```
{
     cacheIdentifier: 'a!a=1'
}
```

You would get following error:

```
Module not found: Error: Can't resolve 'a=1-vue-loader-template"}' in '/Users/haolchen/Documents/__self__/__forked__/vuepress/docs/guide'
 @ ./docs/guide/deploy.md?vue&type=template&id=7ff0bdcd 1:0-324 1:0-324
 @ ./docs/guide/deploy.md
 @ ./lib/app/.temp/routes.js
 @ ./lib/app/app.js
 @ ./lib/app/clientEntry.js
 @ multi ./lib/app/clientEntry.js
```

This PR is to replace all the `!`s in `cacheIdentifier`, and document that `cacheDirectory` path cannot contain `!` (We cannot replace the `!`s in `cacheDirectory`).

refs: https://github.com/vuejs/vuepress/issues/532

## Updated

As Evan's comment, we just `hash` the whole `cacheIdentifier` string.

### DIFF
--- a/lib/loaders/pitcher.js
+++ b/lib/loaders/pitcher.js
@@ -1,5 +1,6 @@
 const qs = require('querystring')
 const loaderUtils = require('loader-utils')
+const hash = require('hash-sum')
 const selfPath = require.resolve('../index')
 const templateLoaderPath = require.resolve('./templateLoader')
 const stylePostLoaderPath = require.resolve('./stylePostLoader')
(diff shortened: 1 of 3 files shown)


lib/loaders/pitcher.js
```diff
@@ -74,7 +74,7 @@ module.exports.pitch = function (remainingRequest) {
     const cacheLoader = cacheDirectory && cacheIdentifier
       ? [`cache-loader?${JSON.stringify({
         cacheDirectory,
-        cacheIdentifier: cacheIdentifier + '-vue-loader-template'
+        cacheIdentifier: cacheIdentifier.replace(/!/g, '') + '-vue-loader-template'
```
yyx990803: I think it's better to just hash the whole string with `hash-sum`, since other chars that may have special meanings in inline requests can also cause issues.

ulivz: Fixed at 448e5b953a649171ce3e73dcabd0f85dbce93931

PR #1345, opened 2018-06-15
compatible with webpack4 mode

wyhaya:
```javascript
{
    mode: 'production'
    // ...
}
```
Using `mode`  in `webpack4` cannot be obtained by `vue-loader`

`vue-loader/lib/index.js` line `45`
```javascript
const isProduction = options.productionMode || minimize || process.env.NODE_ENV === 'production'
```
If you use  `mode` to determine the environment, then `isProduction` cannot catch it. It always is false

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -44,7 +44,10 @@ module.exports = function (source) {
 
   const isServer = target === 'node'
   const isShadow = !!options.shadowMode
-  const isProduction = options.productionMode || minimize || process.env.NODE_ENV === 'production'
+  const isProduction = options.productionMode || 
+                       minimize || 
+                       process.env.NODE_ENV === 'production' ||
+                       loaderContext._compilation.options.mode === 'production'
   const filename = path.basename(resourcePath)
   const context = rootContext || process.cwd()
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -44,7 +44,10 @@ module.exports = function (source) {
 
   const isServer = target === 'node'
   const isShadow = !!options.shadowMode
-  const isProduction = options.productionMode || minimize || process.env.NODE_ENV === 'production'
+  const isProduction = options.productionMode || 
+                       minimize || 
+                       process.env.NODE_ENV === 'production' ||
+                       loaderContext._compilation.options.mode === 'production'
```
yyx990803: is it really safe to access it like this?

sharkykh: Hmm, looks like `_compilation` is deprecated: https://webpack.js.org/api/loaders/#this-_compilation
There's also an issue open on the Webpack repo: https://github.com/webpack/webpack/issues/6496
Maybe it should be done using the [plugin API](https://webpack.js.org/api/compiler-hooks/), if it's possible (not sure).

PR #1368, opened 2018-07-31
Set __file to file basename in production

chrisvfritz:
Combined with [this change](https://github.com/vuejs/vue-component-compiler/commit/541a82483d4ea44b010e2394a65875819a094ad0) in vue-component-compiler, this will allow names to be inferred for _all_ `.vue` files, even those precompiled for production. That means 3rd-party component libraries will never have to define names if they're using `.vue` files! 🙂

We've also made [this change](https://github.com/vuejs/vue-devtools/pull/714) in the devtools, so that we don't allow users to open a file if we only have the basename instead of the full path.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -159,10 +159,15 @@ var component = normalizer(
     code += `\n` + genHotReloadCode(id, hasFunctional, templateRequest)
   }
 
-  // Expose filename. This is used by the devtools and vue runtime warnings.
-  if (!isProduction) {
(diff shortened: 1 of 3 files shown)


privatenumber:
Seems like the docs [here](https://vuejs.org/v2/api/#name) should be updated to reflect this:

> Also, when inspecting an app in the vue-devtools, unnamed components will show up as <AnonymousComponent>, which isn’t very informative.

PR #1387, opened 2018-08-21
Fix the snapshot testing issue on windows

elonmitchell:
The back slash should be replaced with the slash for consistent hash.


elonmitchell:
@yyx990803 can you please take a look this asap? My project already made a huge progress with the vue-loader v13 so can't upgrade to v15. Thanks for your help in advance.

yyx990803:
Published 13.7.3.

Issue #516, opened 2016-12-02
Path of Vue component file is exposed by `__vue_options__.__file`.


songsiqi:
I use vue-loader@^9.

```
__vue_options__.__file = "/Users/username/Documents/path/to/component.vue"
__vue_options__.render = __vue_template__.render
__vue_options__.staticRenderFns = __vue_template__.staticRenderFns
```

How to avoid this?

LinusBorg:
This is [disabled in production](https://github.com/vuejs/vue-loader/blob/912ec9c14cf7b44b03fdfaabf7995b4c5f149350/lib/loader.js#L257) and is used in development to provide better error messages.

fritx:
Excuse me, how to turn that `minimize` on ???

kevingermain:
I'm joining fritx :)

LinusBorg:
Just minimize your code, e.g.  with uglify webpack plugin. We already do or in our template.

justblender:
I'm sorry for bumping this old issue, but is there any reason why even production builds now expose component's filename? Maybe there should be a way to disable this behavior?

LinusBorg:
See:

https://github.com/vuejs/vue-loader/blob/f0beed3296d676dfdfcad85f115c9e75aaaca08b/lib/index.js#L178-L186

LinusBorg:
you can always 
* Fork vue-loader
* Remove the code you don't like.

If you actually want to make an effort to have this changed in vue-loader itself, please open a feature request with an explanation about why this is a problem for you and why offering a way to remove it is valuable to you, and if you want to help us even more, send a PR.

gitlab-winnie:
I have created https://github.com/vuejs/vue-loader/pull/1446 to document the behavior.

2-5:
For those who inspected their minified bundles and were most surprised by this sensitive information leak, here is how to fix it if you use `webpack`. We are going to go over the `vue-loader` output and 
strip it with a regular expression, with the help of the `string-replace-loader` plugin:

```
npm install --save-dev string-replace-loader
```

And in your webpack config:

```javascript
module: {
  rules: [
    {
      test: /\.vue$/,
      use: [
        {
          loader: 'string-replace-loader',
          options: {
            search: '\\.options\\.__file = ".*"',
            replace: '.options.__file = "a.vue"',
            flags: '',
          },
        },
        { loader: 'vue-loader' },
      ],
    },
  ],
},
```

[...]

PR #1446, opened 2018-12-04
Document __file option of imported components

gitlab-winnie:
Add missing documentation for the __file option that is exposed by vue-loader

see https://github.com/vuejs/vue-loader/blob/v15.4.0/lib/index.js#L173-L181

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -126,3 +126,6 @@ Contributions for other editors/IDEs are highly appreciated! If you are not usin
 
 Inside each block you shall use the comment syntax of the language being used (HTML, CSS, JavaScript, Jade, etc). For top-level comments, use HTML comment syntax: `<!-- comment contents here -->`
 
+## `__file` Option
+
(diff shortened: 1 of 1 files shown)


gitlab-winnie:
@chrisvfritz Can you please take a look? 😃

LinusBorg:
Theos is not a public API

gitlab-winnie:
@LinusBorg  Thanks for your comment! 👍

Based on the inline comment in the code and on the discussion in https://github.com/vuejs/vue-loader/issues/516 it is intentionally exposed in development and production environment though—which is quite public. 😃

Is there any chance to make this a public API other than documenting it?

gitlab-winnie:
> We might break it with a patch release.

...and that's what I wanted to sneakily change here. Unfortunately you caught me—what a bummer. 😞 

> I mean that this is an API that we don't intend for people to use for anything.

Now my problem is that I would like to use the API. Do you have any suggestion for me going forward? Are these my only options:

- risk to have things broken with a patch release update of vue-loader
- maintain a fork of vue-loader

PR #1469, opened 2019-01-08
feat: support webpack 5 hooks

clarkdo:


lib/plugin.js
```diff
@@ -10,7 +10,12 @@ class VueLoaderPlugin {
     if (compiler.hooks) {
       // webpack 4
       compiler.hooks.compilation.tap(id, compilation => {
-        compilation.hooks.normalModuleLoader.tap(id, loaderContext => {
+        let { normalModuleLoader } = compilation.hooks;
+        if (!normalModuleLoader) {
```
sodatea: `compilation.hooks.normalModuleLoader` still exists in webpack 5, just deprecated, and a `get` operation will trigger the warning, so it's better to let the new `getCompilationHooks` method take a higher priority.

package.json
```diff
@@ -34,7 +34,7 @@
   },
   "peerDependencies": {
     "css-loader": "*",
-    "webpack": "^4.1.0"
+    "webpack": "^4.1.0 || ^5.0.0"
```
sodatea: To support alpha & beta versions, we need to set the range as `^5.0.0-0`

clarkdo:
@sodatea Thanks for the review, changed as your suggestion 😄

PR #1471, opened 2019-01-11
docs(zh): updated eslint plugin link

Jinjiang:
Just as the same to en.
Thanks.


Jinjiang:
@sodatea not sure. Just follow the en docs: https://github.com/vuejs/vue-loader/blob/master/docs/guide/linting.md
But you are right. I will update this link for all languages.
Thanks.

PR #1511, opened 2019-03-10
fix: id should only depend on file path

clyfish:
Why does `id` depend on `source`?

If `index.js` dynamic imports `a.vue`, `script` in `a.vue` is changed and `style` in `a.vue` is not changed,
the compiled js from `index.js` and compiled css from `a.vue` should not be changed.

But when `id` depend on `source`, the compiled js from `index.js` and compiled css from `a.vue` are changed.


yyx990803:
Because it's possible for two components (from different projects) to have the same file path but different contents. If a user compiles their component and publishes it, and then used by another user, the CSS scoping can potentially conflict.

PR #1519, opened 2019-03-19
 支持自动导入vue文件同级同名的js文件

ChangedenCZD:
如果vue文件同级目录中存在同名的js文件，则会自动导入该js文件作为vue组件的脚本。
如以下目录结构：
```
|--myComponent
  |--demo.vue
  |--demo.js
```
demo.vue在编译时，会自动导入demo.js并替换掉vue文件中的<script>标签内容。

[详细例子](https://github.com/ChangedenCZD/vue-loader/tree/dev-chan/example)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ example/dist
 docs/.vuepress/dist
 test/.cache
 .idea
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 example/dist
 docs/.vuepress/dist
 test/.cache
+.idea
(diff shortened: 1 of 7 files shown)


Justineo:
Thank you. But I don't think we will accept this PR.

1. Theoretically, this would be a breaking change.
2. This would make the dependency less explicit.

ChangedenCZD:
> Thank you. But I don't think we will accept this PR.
> 
> 1. This would be a breaking change.
> 2. This would make the dependency less explicit.

因此我希望可以通过配置文件进行设置，目前没找到相关权限。
该功能解决的是某些情景下由于业务代码过多造成维护困难的问题。

sodatea:
Why not just use `import` in `.vue` file?

Issue #1454, opened 2018-12-25
Not able to use a compiler proxy


lucasmpaim:
### Version
15.4.2

### Reproduction link
[ https://github.com/lucasmpaim/testbuild]( https://github.com/lucasmpaim/testbuild)

### Steps to reproduce
run yarn build

### What is expected?
The build has compiled without problem

### What is actually happening?
raised a error compiler.parseComponent is not a function

---
Using this solution: https://github.com/vuejs/vue/issues/3934 with vue-cli project

Using this configuration: 
```javacsript
  chainWebpack: config => {
    config.module.rule('vue')
      .use('vue-loader')
      .loader('vue-loader')
      .tap(options => {
        options.compiler = VueTemplateCompilerProxy
        options.compilerOptions.preserveWhitespace = false;
        return options
      })
  }
```

on build process I receive the exception: `compiler.parseComponent is not a function`, In development is working perfectly, in `vue-loader/lib/index.js` I put an `console.log` to check why this is happening, and I notice in sometimes the compiler is a blank object `{}`.
The log:
```
App.vue { parseComponent: [Function: parseComponent],
  compile: [Function: compile],
  compileToFunctions: [Function: compileToFunctions],
  ssrCompile: [Function: compile],
  ssrCompileToFunctions: [Function: compileToFunctions] }
App.vue { parseComponent: [Function: parseComponent],
  compile: [Function: compile],
  compileToFunctions: [Function: compileToFunctions],
  ssrCompile: [Function: compile],
  ssrCompileToFunctions: [Function: compileToFunctions] }
App.vue { parseComponent: [Function: parseComponent],
  compile: [Function: compile],
  compileToFunctions: [Function: compileToFunctions],
  ssrCompile: [Function: compile],
  ssrCompileToFunctions: [Function: compileToFunctions] }
App.vue {}
.....
```

Adding:
```javascript
   if(options.compiler && !Object.keys(options.compiler).length) options.compiler = null
```
Before parser executes, everting work fine.

viruscamp:
I met the same problem. I have temporarily fixed it by disable [parallel](https://cli.vuejs.org/config/#parallel) in vue.config.js.
I think it is caused by the use of 'thread-loader', which is enabled default by vue-cli.

The main node process passes stringified  webpack options (includes vue-loader options) to worker processes. Functions could not be passed.

PR #1533, opened 2019-04-04
feat: Add compilerFile for options

vincenting:
For some reasons, we alway get a error when building in production mode: `compiler.parseComponent is not a function`.

After we tried to debug it, we found some loaders call vue-loader with stringified args, which resulting in losing all functions we declared in our config. (`VueTemplateCompiler`)

Here is another similar issue https://github.com/vuejs/vue-loader/issues/1454

The easiest way to fix it is using a string param to pass the configuration of compiler. So, use a compilerFile: string will be a choose for this case.

```js
{
  compilerFile: 'absPathToCompilerFile',
} 
```


vincenting:
https://github.com/webpack-contrib/terser-webpack-plugin#parallel

In production mode, `parallel` is enabled by default, which as a result leading to this issue. We are now disable `parallel` to avoid this error in production.

`vue-loader` is now using `loader-utils#getOptions ` to parse params. Which means all the data structure supported by JSON is suggested as options. `VueTemplateCompiler` is a kind of data structure with Function, which will leave some risks in some certain conditions.

viruscamp:
1. It's better to use the same option name.
compiler:  String | VueTemplateCompiler
2. Other loader-options that cannot be stringified , should be treat the same way. Just like:
{
  compiler:  String | VueTemplateCompiler
  compilerOptions: {
    modules: Array<string | ModuleOptions>,
    directives: { [key: string]: String | Function }
  }
}

PR #1535, opened 2019-04-06
Template Specification

znck:
[preview](https://github.com/znck/vue-loader/blob/template-spec/docs/spec/template.md)


octref:
That's a good start. This would be useful for eslint-plugin-vue parser, etc, but I'm more interested in spec for the sections. For example, is `<template lang="html">` allowed? How about `<script lang="babel">`? Should the syntax inside `<style>` be postcss, or `<style>` is reserved for CSS and people should use `<style lang="postcss">` for postcss?

The current spec is pretty loose: https://vue-loader.vuejs.org/spec.html#vue-single-file-component-sfc-spec.

yyx990803:
I like this! This would definitely be nice to have for tooling authors and also for ourselves (to discover and formally resolve edge cases).

One improvement we can do is to use a more formal grammar description along the lines of [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form). The ECMAScript spec uses something similar: https://www.ecma-international.org/ecma-262/9.0/index.html#sec-names-and-keywords

Screen Shot 2019-04-08 at 9 18 30 AM

There are also parser-generator tools that can take a BNF-like grammar and automatically generate parsers, e.g. [antlr](https://github.com/antlr), [jison](https://github.com/zaach/jison) (a JavaScript port of [bison](http://dinosaur.compilertools.net/bison/index.html)). The [Antlr v4 HTML grammar files](https://github.com/antlr/grammars-v4/tree/master/html) can be a good source of reference.

yyx990803:
Also, like @octref said, we also want a more well-defined version of the SFC spec. However, I think the SFC spec and the Vue template spec should be separate. The SFC spec should live inside [component-compiler-utils](https://github.com/vuejs/component-compiler-utils), and the template spec should live inside Vue core repo.

octref:
Just today I run into an issue where spec for template would help: https://github.com/vuejs/vetur/issues/1189. Would unquoted attributes be valid for Vue templates?

yyx990803:
Re unquoted attributes: should be consistent with HTML. i.e. `v-if=ok` is valid. It's not a recommended style but is indeed allowed.

docs/spec/template.md
```diff
@@ -0,0 +1,83 @@
...
+
+##### Dynamic Directive Argument
+
+The dynamic directive argument must include following parts:
+
+- The string `[`
+- A JavaScript expression which must not include following characters:
+  - `\t` (tab)
+  - `\r` (carriage return)
+  - `\n` (new lint)
```
jamesgeorge007:
```suggestion
  - `\n` (new line)
```

PR #1542, opened 2019-04-23
types: add loader options declarations

kbtzr:
This PR essentially adds the interface `VueLoaderOptions` that can aid type checking
when writting Webpack configuration with Typescript:

```ts
// webpack.config.ts

import { VueLoaderOptions, VueLoaderPlugin } from 'vue-loader'

export default {
  module: {
    rules: {
      // ...
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          // * autocompletion available here *
        } as VueLoaderOptions
      }
    }
  },
  plugins: [
    new VueLoaderPlugin()
  ]
}
```

I'm studying the possibility of using a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#Discriminated-Unions) to assign the `options` type when a match `loader: 'vue-loader'` is found, but this will come in a later PR. For now the casting on the options object is required.

---

Please note the compiler interface use types that were added only in Vue 2.6 (vuejs/vue#7918), but folks with older versions of the template compiler shouldn't experience any errors, they will just not get any type checking on `compile` and `compileOptions` properties.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,7 +1,40 @@
 import { Plugin } from 'webpack'
+import {
+  CompilerOptions,
+  parseComponent,
+  compile,
+  compileToFunctions,
+  ssrCompile,
+  ssrCompileToFunctions,
+  generateCodeFrame
+} from 'vue-template-compiler'
 
 declare namespace VueLoader {
   class VueLoaderPlugin extends Plugin {}
+
+  interface VueLoaderTemplateCompiler {
+    parseComponent?: typeof parseComponent
+    compile?: typeof compile
+    compileToFunctions?: typeof compileToFunctions
+    ssrCompile?: typeof ssrCompile
+    ssrCompileToFunctions?: typeof ssrCompileToFunctions
+    generateCodeFrame?: typeof generateCodeFrame
+  }
+
+  interface VueLoaderOptions {
+    transformAssetUrls?: { [tag: string]: string | Array<string> }
+    compiler?: VueLoaderTemplateCompiler
+    compilerOptions?: CompilerOptions
+    transpileOptions?: Object
+    optimizeSSR?: boolean
+    hotReload?: boolean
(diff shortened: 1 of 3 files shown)


lib/index.d.ts
```diff
@@ -1,7 +1,40 @@
...
+    compile?: typeof compile
+    compileToFunctions?: typeof compileToFunctions
+    ssrCompile?: typeof ssrCompile
+    ssrCompileToFunctions?: typeof ssrCompileToFunctions
+    generateCodeFrame?: typeof generateCodeFrame
+  }
+
+  interface VueLoaderOptions {
+    transformAssetUrls?: { [tag: string]: string | Array<string> }
+    compiler?: VueLoaderTemplateCompiler
```
sodatea: `VueTemplateCompiler` type can be imported from `@vue/component-compiler-utils`.
https://github.com/vuejs/component-compiler-utils/blob/442d7b221d2075b2b53d8f6d4e1a4b4e6ab911e3/lib/types.ts#L16-L28

kbtzr: Thanks! I noticed a mismatch between the compiler options from [component-compiler-utils](https://github.com/vuejs/component-compiler-utils/blob/442d7b221d2075b2b53d8f6d4e1a4b4e6ab911e3/lib/types.ts#L33) and from [vue-template-compiler](https://github.com/ktsn/vue/blob/6449085949305eefad08402cbf7905f2e67286ff/packages/vue-template-compiler/types/index.d.ts#L6). Which one you think is more appropriate for me to use?

sodatea: The latest one in the core repo is more complete: https://github.com/vuejs/vue/blob/29e745f99f43d0629df547da80b016996ae70872/packages/vue-template-compiler/types/index.d.ts#L6

Both are fine, though. As all the fields are optional.

PR #1560, opened 2019-06-21
fix getting shadow root when component is functional

zrh122:
Add fetching `$root` from parent because functional components' context doesn't have the `$root` property. It will close [issue](https://github.com/vuejs/vue-cli/issues/4097).


Marcus-Rise:
Still waiting for accepting PR from @zrh122

PR #1569, opened 2019-07-16
fix: use "api.isRecorded" instead of "module.hot.data"

liximomo:
We should not use `module.hot.data`, which is highly non-reliable. Plugins like [lazy-compile-webpack-plugin](https://github.com/liximomo/lazy-compile-webpack-plugin) use a dummy loader to prevent webpack from building async modules, the module will be compiled and hot-update on demand.  In such a case, the `module.hot.data` is truthy but the record does not exist, which will cause a running error in `api.reload`. I think `api.isRecorded` is more reliable and reasonable.

### DIFF
--- a/lib/codegen/hotReload.js
+++ b/lib/codegen/hotReload.js
@@ -19,7 +19,7 @@ if (module.hot) {
   api.install(require('vue'))
   if (api.compatible) {
     module.hot.accept()
-    if (!module.hot.data) {
+    if (!api.isRecorded('${id}')) {
       api.createRecord('${id}', component.options)
     } else {
       api.${functional ? 'rerender' : 'reload'}('${id}', component.options)


lib/codegen/hotReload.js
```diff
@@ -19,7 +19,7 @@ if (module.hot) {
   api.install(require('vue'))
   if (api.compatible) {
     module.hot.accept()
-    if (!module.hot.data) {
+    if (!api.isRecorded('${id}')) {
```
yyx990803: could just be `api.isRecorded(id)`

liximomo: Sorry, couldn't get it. Did you mean to remove the `${}`? I don't think that could work.

yyx990803: `id` is already a string isn't it?

liximomo:
Yes, but we are in a template string. You can see the codes below this line
also does the same thing.

Evan You <notifications@github.com> 于 2019年7月17日周三 11:44写道：

lib/codegen/hotReload.js
liximomo: Yes, but we are in a template string. You can see the codes below this line also does the same thing. Let me know if I miss anything.

yyx990803:
Oh nvm... Haha I was looking at the diff without syntax highlighting

liximomo:
FYI. This PR may also solve this problem.

https://stackoverflow.com/questions/52339206/vuejs-hot-reload-cannot-read-property-ctor-of-undefined.

sodatea:
Though I'm still struggling to understand why `module.hot.data` is unreliable (is it because there may be multiple sources of `dispose` call?), this commit seems to work great.

liximomo:
@sodatea Hope these pseudo-codes can provide some insight

```
compiler.hooks.compilation.tap('LazyCompilePlugin', compilation => {
  compilation.hooks.buildModule.tap('LazyCompilePlugin', wpModule => {
    if (ShouldBlock) {
      // relace the whole loaders
      wpModule.loaders = [
        {
          loader: dummy,
        },
      ];
    } else {
      // restore original loaders
      // The same module will be hot-updated, and module.hot.data will be true.
      // Since vue-loader has't run before, there is no any records.
    }
  });
});

```

liximomo:
🎉

PR #1615, opened 2019-12-03
docs: Guide to use `::v-deep`

cia-rana:
Deep Selectors using `/deep/` will no longer work with the latest `vue-loader` that depend on `"@vue/component-compiler-utils": "^3.0.0`, so guide users to use v-deep.


sodatea:
It still works with `node-sass`. It just doesn't work with dart `sass` (which is used by `rollup-plugin-vue` but in `vue-loader` it's up to the user).

PR #1616, opened 2019-12-10
Update @vue/component-compiler-utils to 3.1.0

NE-SmallTown:


NE-SmallTown:
@sodatea friendly ping~

sodatea:
Well, we don't typically update minor versions in a PR because `^3.0.0` will automatically resolve to the latest version…
So we only do the version maintenance in a batch periodically.

But thanks anyway.

PR #1627, opened 2020-01-12
chore(deps): migrate to memfs

jamesgeorge007:
`memory-fs` is deprecated, switch over to `memfs`.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -76,6 +76,7 @@ function bundle (options, cb, wontThrowError) {
 
   const webpackCompiler = webpack(config)
   webpackCompiler.outputFileSystem = mfs
+  webpackCompiler.outputFileSystem.join = path.join.bind(path)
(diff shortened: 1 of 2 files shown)


test/utils.js
```diff
@@ -4,9 +4,9 @@ const hash = require('hash-sum')
 const { JSDOM, VirtualConsole } = require('jsdom')
 const webpack = require('webpack')
 const merge = require('webpack-merge')
-const MemoryFS = require('memory-fs')
+const { createFsFromVolume, Volume } = require('memfs')
 
-const mfs = new MemoryFS()
+const mfs = new createFsFromVolume(new Volume()) // eslint-disable-line
```
sodatea: Isn't the `new` before `createFsFromVolume` redundant?

jamesgeorge007: #1633

PR #1635, opened 2020-02-04
Is the test option of template compiler rule required?

Senasiko:
If the `test` option  of template compiler rule  is removed, some files  like `template.html?vue&type=template` can match the rule, this is expected. So I think the `test` option is not required. 
Also there has some causes that have to use `test`?


meteorlxy:
This PR can be closed as #1673 has been merged

PR #1646, opened 2020-03-02
chore: remove temporary unused code

unixzii:
Maybe this is trivial, but removing that makes the code looks better :)


sodatea:
Thanks!

PR #1649, opened 2020-03-12
fix minor problems

bichikim:
1. add types
2. the `rm` command is not working in Windows OS (so use clean-webpack-plugin Instead)
3. 
```
if(a === 'b') {
  return true
}
return false
```
can be
```
return a === 'b'
```

4. `\/|\\|@` can be `([\/\\@])`

5. the Loader can be 'string' also

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "dev": "tsc --watch",
     "build": "tsc",
     "dev-example": "webpack-dev-server --config example/webpack.config.js --inline --hot",
-    "build-example": "rm -rf example/dist && webpack --config example/webpack.config.js --env.prod",
+    "build-example": "webpack --config example/webpack.config.js --env.prod",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "dev": "tsc --watch",
     "build": "tsc",
     "dev-example": "webpack-dev-server --config example/webpack.config.js --inline --hot",
-    "build-example": "rm -rf example/dist && webpack --config example/webpack.config.js --env.prod",
```
posva: We could simply use [rimraf](https://www.npmjs.com/package/rimraf) without adding a webpack plugin

bichikim: If so the build-example script should be 
`rimraf -rf example/dist && webpack --config example/webpack.config.js --env.prod`

PR #1665, opened 2020-04-17
bump(deps): bump @vue/compiler-sfc and vue to ^3.0.0-beta.1

pikax:
Bump to `beta`, otherwise you get:

```
Error: vue-loader requires @vue/compiler-sfc to be present in the dependency tree.
    at Object.<anonymous> (.\node_modules\vue-loader\dist\index.js:10:11)
```


sodatea:
I don't understand.
These are `devDependencies`, shouldn't lead to any errors in production. `3.0.0-beta.1` should also have been covered by the range `^3.0.0-alpha.1`. You can test it on https://semver.npmjs.com/.

pikax:
I thought it was weird, just checked against the `vue-next-webpack-preview` and it is working without this :/

Probably got confused 😄

PR #1668, opened 2020-04-28
feat:custom block loader support id query parameter

iptop:
In some cases ,custom block loader You may need an id parameter,If you do not modify vue-loader,The custom block loader implementation seems to get very complicated


iptop:
@sodatea The unit test was considered timed out for some reason and the pull request was resubmitted

PR #1674, opened 2020-05-09
refactor!: replace module.exports with ES module export

### LINKED ISSUE #1667: No named VueLoaderPlugin export in TypeScript webpack config
### Version
16.0.0-alpha.3

### Reproduction link
File of interest:
[https://github.com/davelsan/vue-next-starter/blob/master/webpack.config.ts](https://github.com/davelsan/vue-next-starter/blob/master/webpack.config.ts)
Full repo:
[https://github.com/davelsan/vue-next-starter](https://github.com/davelsan/vue-next-starter)

### Steps to reproduce
Simple version:
1. Create a `webpack.config.ts` file
2. Try to `import { VueLoaderPlugin } from 'vue-loader'`

Using the provided repro
1. Comment `lines [13-15]` in `webpack.config.ts`

### What is expected?
There should be a named export or a type definition for the default export that allows creating a `new VueLoaderPlugin()` object.

> webpack.config.ts:8:10 - error TS2305: Module '"./node_modules/vue-loader/dist"' has no exported member 'VueLoaderPlugin'.

Additional Comments
---
The [repro](https://github.com/davelsan/vue-next-starter) repository is a working Vue application built without the CLI toolchain. To work around the errors listed above, I augmented the 'vue-loader' plugin as described in the TypeScript [documentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation):

```ts
declare module 'vue-loader' {
  export class VueLoaderPlugin extends Plugin { }
}
```

This is only an issue when using a `webpack.config.ts`, the [js version](https://github.com/davelsan/vue-next-starter/blob/master/___webpack.config.bak.js) works fine.

sodatea:
This change fixes the emitted TypeScript type definitions

Closes #1667.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,7 +24,10 @@ import { selectBlock } from './select'
 import { genHotReloadCode } from './hotReload'
 import { genCSSModulesCode } from './cssModules'
 import { formatError } from './formatError'
+
 import VueLoaderPlugin from './plugin'
+
+export { VueLoaderPlugin }
 
 export interface VueLoaderOptions {
   transformAssetUrls?: SFCTemplateCompileOptions['transformAssetUrls']
@@ -37,7 +40,10 @@ export interface VueLoaderOptions {
 
 let errorEmitted = false
 
-function loader(this: webpack.loader.LoaderContext, source: string) {
+export default function loader(
+  this: webpack.loader.LoaderContext,
+  source: string
+) {
   const loaderContext = this
 
   // check if plugin is installed
@@ -244,6 +250,3 @@ function attrsToQuery(attrs: SFCBlock['attrs'], langFallback?: string): string {
   }
   return query
 }
-
-export { VueLoaderPlugin }
-export default loader
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,8 +24,7 @@ import { selectBlock } from './select'
 import { genHotReloadCode } from './hotReload'
 import { genCSSModulesCode } from './cssModules'
 import { formatError } from './formatError'
-
-const VueLoaderPlugin = require('./plugin')
+import VueLoaderPlugin from './plugin'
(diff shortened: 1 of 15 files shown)


src/index.ts
```diff
@@ -38,7 +40,10 @@ export interface VueLoaderOptions {
 
 let errorEmitted = false
 
-const loader: webpack.loader.Loader = function(source: string) {
+export default function loader(
```
yyx990803, line 43: This would require `require('vue-loader').default`? Does webpack have built-in esModule default interop?

test/utils.ts
```diff
@@ -0,0 +1,133 @@
...
+      // this isn't technically needed, since the default `vue` entry for bundlers
+      // is a simple `export * from '@vue/runtime-dom`. However having this
+      // extra re-export somehow causes webpack to always invalidate the module
+      // on the first HMR update and causes the page to reload.
+      vue: '@vue/runtime-dom'
+    }
+  },
+  resolveLoader: {
+    alias: {
+      'vue-loader': require.resolve('../src')
```
yyx990803, line 28: Does this work because jest registers require hooks and compiles the ts src for all require calls? I think we should probably run build before test and run the test on the build dist files since that is closer to production.

src/index.ts
sodatea: Yeah, webpack uses `loader-runner` to run loaders, and it supports `.default` https://github.com/webpack/loader-runner/blob/master/lib/loadLoader.js#L26

Or we can provide another `.cjs` entry if compatibility is a concern.
But I don't think it's necessary, as `vue-loader` is only supposed to be called by webpack.

test/utils.ts
sodatea: Done

PR #1678, opened 2020-05-25
fix(next): use `export = xxx` instead of `exports.xxx`

underfin:
fix https://github.com/vuejs/vue-next/issues/1229

## Bug Reson
webpack import module by node solution and caused `pitcher` isn't loaded. 
This can be fix by node `exports.xxx`  === ts `export = xxx`.

### DIFF
--- a/src/pitcher.ts
+++ b/src/pitcher.ts
@@ -125,4 +125,4 @@ function shouldIgnoreCustomBlock(loaders: Loader[]) {
   return actualLoaders.length === 0
 }
 
-export default pitcher
+export = pitcher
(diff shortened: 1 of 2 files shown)


sodatea:
Fixed in https://github.com/vuejs/vue-loader/commit/79a94dda3f2c89755ac21c4555f53b13111452bf

I chose to use ES modules to keep consistent with other modules in the repo.

The export in `plugin.ts` does not matter because the public API should be `import { VueLoaderPlugin } from 'vue-loader'` or `const { VueLoaderPlugin } = require('vue-loader')`.

PR #1680, opened 2020-05-26
fix: skip matching rules with 'enforce'

Jamie-Yang:
skip matching rules with 'enforce'.
eg. the rule of eslint-loader is before the rule of vue-loader, will throw this error: '[VueLoaderPlugin Error] No matching rule for .vue files found.'

### DIFF
--- a/src/pluginWebpack5.ts
+++ b/src/pluginWebpack5.ts
@@ -96,7 +96,10 @@ class VueLoaderPlugin implements Plugin {
     let vueRules: Effect[] = []
 
     for (const rawRule of rules) {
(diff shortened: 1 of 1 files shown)


sodatea:
Do you have a reproduction repo for the described issue?
I can't get the error on my local project.

Jamie-Yang:
Sure. Check out my vue 3 boilerplate project. https://github.com/Jamie-Yang/vue3-boilerplate
Line 69 to line 84 of ./service/config/base.js

sodatea:
Thanks! I'm able to reproduce it now.

PR #1706, opened 2020-07-31
fix: Fix id differences between different operating systems

denghui1010:
Id will affect the hash calculation of webpack, thus affecting the final cache


sodatea (approved):
Thanks!

PR #1714, opened 2020-08-12
docs: 'prependData' option renamed to 'additionalData' in sass-loader 9

ux-engineer:
Option `prependData `should be renamed `additionalData` in `vue.config.js` as per [sass-loader v9.0.0 changelog](https://github.com/webpack-contrib/sass-loader/releases/tag/v9.0.0) and corresponding [doc](https://github.com/webpack-contrib/sass-loader#additionaldata).


sodatea (approved):
Thanks!

PR #1730, opened 2020-09-26
improv: adding option for hmr when target is node

shubhamzanwar:


shubhamzanwar:
When creating native apps with the new renderer package, the target is node and hence, we need some sort of mechanism to enable hot reloading.

shubhamzanwar:
Whoops wrong branch. Should probably be on the `next` branch

PR #1731, opened 2020-09-26
improv: adding option for hmr when target is node

shubhamzanwar:


shubhamzanwar:
When creating native apps with the new renderer package, the target is node and hence, we need some sort of mechanism to enable hot reloading.

shubhamzanwar:
Could we please get some attention to this PR? :(

sodatea:
Does this feature solve your problem? https://github.com/vuejs/vue-loader/pull/1764
If I understand correctly, you are targeting your components to a custom renderer, so you might not only want to enable HMR, but also avoid all other SSR-specific optimizations.

PR #1735, opened 2020-10-01
fix: fix script bindings should not be a  undefined value

PeckZeg:
when bindings pass as a query string, should not be `"undefined"` string value.


PeckZeg:
timeout in unit test - "add custom blocks to the webpack output"?

sodatea (approved):
Thanks!

PR #1745, opened 2020-10-22
fix: set default render when template is empty

meteorlxy:
`@vue/compiler-sfc` will return `null` if `<template>` is empty, so we need default render to handle it.

https://github.com/vuejs/vue-next/blob/f4621ff5ee4abe924d985177956af3ddc9bb378f/packages/compiler-sfc/__tests__/parse.spec.ts#L114-L119

Also, this is also handled by `rollup-plugin-vue`:

https://github.com/vuejs/rollup-plugin-vue/blob/6960203fdc4285b4322a2c4f061f4a2773d49c9c/src/index.ts#L469

Without this default render, it would cause "Component is missing template or render function" warning

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -151,6 +151,7 @@ export default function loader(
 
   // template
   const renderFnName = isServer ? `ssrRender` : `render`
+  const renderReplace = `script.${renderFnName} = ${renderFnName}`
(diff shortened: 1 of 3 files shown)


meteorlxy:
Close due to https://github.com/vuejs/vue-next/issues/2463

PR #1756, opened 2020-10-30
Improve error message

### LINKED ISSUE #1694: Add specific error message when there is no default export for a Component
When a component `.vue` file does not have a default export, the `Component` cannot be loaded (as in #1234). Currently, the error message is unclear (I got "Uncaught TypeError: Cannot set property 'render' of undefined"). It would be clearer to generate an error saying explicitly that the component definition did not have a default export which is required.

`Error: Component MyComponent.vue does not have a default export. See https://vuejs.org/v2/guide/components-registration.html#Local-Registration .`

yamacent:
close #1694


sodatea (changes requested):
Thanks for the help! But could you add a test case for this feature? Because if it's a compile-time error, I doubt such code would have any effect.

yamacent:
@sodatea 
Thanks for the comment!
This PR tries to replace the runtime error message `Uncaught TypeError: Cannot set property 'render' of undefined`.
Do you think it should be compile-time error? And can that realize with small change?

PR #1768, opened 2020-12-10
Namespace module ids with package.json name. Resolves #1767.

### LINKED ISSUE #1767: Namespace module ids for different projects
### Version
16.0.0-beta.7

### Reproduction link
[https://github.com/joeldenning/vue-loader-hmr-bug](https://github.com/joeldenning/vue-loader-hmr-bug)

### Steps to reproduce
The steps are in the README of the Github repo.

### What is expected?
Both apps will mount.

One of the apps is mounted twice, and the other is not mounted at all.

---
The microfrontends community often runs multiple Vue CLI projects in a single web app. I help maintain single-spa and this was originally reported to me in https://github.com/single-spa/single-spa/issues/672.

This bug occurs because vue-loader generates the same module id for files in different projects. Since both app1 and app2 have `src/App.vue` files, they have the same module ids.

See [this code](https://github.com/vuejs/vue-loader/blob/decb4dd031247460fd26132f9de3962982926357/src/index.ts#L114) which shows that the module id is based entirely on the relative file path of the vue component. Since Vue CLI scaffolds projects with the same file paths, this results in collisions.

### Proposed fix:

Namespace the module ids based on the package.json name, current working directory, or some other field that vue-loader already has available to it. This would ensure that `src/App.vue` would result in a different module id in two different projects. I'm happy to put together a fix for this, but am unsure what would be best to use as the namespace. Am leaning towards package.json name.

joeldenning:
See #1767 for explanation. I'm looking forward to any feedback on a better way to get the package.json name, or a better namespace to use entirely rather than the package.json name.


joeldenning:
I created this pull request to the wrong branch, and my code was also branched off of master instead of `next`. That's why the diff got so weird here. Github is not letting me change the base branch of this pull request anymore, so I will create a new PR for this.

joeldenning:
I've created https://github.com/vuejs/vue-loader/pull/1769 which has the correct base branch.

PR #1774, opened 2020-12-20
fix: compatible with webpack5 externals script

fangbinwei:
fixes https://github.com/vuejs/vue-cli/issues/6149

I have added test to verify that `api.install(Vue)` will be executed correctly


lib/index.js
```diff
@@ -176,7 +176,11 @@ var component = normalizer(
   }
 
   if (needsHotReload) {
-    code += `\n` + genHotReloadCode(id, hasFunctional, templateRequest)
+    code = [
+      `import Vue from 'vue'`,
```
sodatea, line 180: Does it have to be `import`? I'm not sure if it would break some CommonJS code.

fangbinwei: > if it would break some CommonJS code.

I'm not sure what 'some CommonJS code' means here

As far as I know,  `import` in webpack can handle both ES Module and CommonJS
 
```js
import Vue from 'vue/dist/vue.esm.js'
console.log(Vue.version) // works
```
```js
import Vue from 'vue/dist/vue.common.js'
console.log(Vue.version) //works
```

but `require` is different
```js
const Vue = require('vue/dist/vue.esm.js')
console.log(Vue.default.version)
```
```js
const Vue = require('vue/dist/vue.common.js')
console.log(Vue.version)
```

So [`Vue = vue.__esModule ? vue.default : vue`](https://github.com/vuejs/vue-hot-reload-api/blob/eb51dc7e8bffc5b23e3df8b963924e1fc2bfd881/src/index.js#L15) in `vue-hot-reload-api` is necessary for `require('vue')`.  But it's redundant for `import Vue from 'vue'`, since webpack will check `__esModule` in generated code and import Vue as expected.

Therefore, I think `import Vue form 'vue'` won't break codes below
```js
var api = require(`path_to_vue-hot-reload-api`)// or import api from 'path_to_vue-hot-reload-api'
api.install(...)
```


> Does it have to be import?

From the documentation, [`externalsType: 'script'`](https://webpack.js.org/configuration/externals/#example) works fine with `import`.  

And it can work with `require` like example below. Promise fulfilled after the script is loaded. 
```js
require('vue').then(Vue => {
  
})
```
When using `import Vue from 'vue'` webpack will  handle the Promise, I think it's because ESModule can be statically analyzed. Compared to `require`, webpack can do more for `import` at the compile stage.

I didn't see `externalsType: script` related discussions, maybe I can get more information from source code

sodatea: Try this (in `HelloWorld.vue`):

```js
import Vue from 'vue' // eslint-disable-line

module.exports = {
  name: 'HelloWorld',
  props: {
    msg: String
  }
}
```

vs

```js
module.exports = {
  name: 'HelloWorld',
  props: {
    msg: String
  }
}
```

fangbinwei: Maybe the import statement added by `vue-loader-v15` doesn't conflict with the sciprt in `HelloWorld.vue`?

image


some problem I find

- webpack will generated some code to `await` the script tag resolve, it doesn't compatible with ie11 (Need to investigate)
- `vue-hot-reload-api.install/compatible` is a little outdated

sodatea: Oh sorry, I forgot it is in another module rather than the script module…

xuemanchi:
Any news？

PR #1802, opened 2021-03-04
fix: add __moduleIdentifier for ssr _registeredComponents

asthesky:
hi, is there any solution for SSR preload async components? maybe I didn't find it, and i find an crude method like this ,
 ```
       const { __moduleIdentifier } = this.$options;
      if (__moduleIdentifier) {
        const ctx = useSSRContext();
        ctx._registeredComponents.add(__moduleIdentifier);
      }
 ```
 but this need add a moduleIdentifier, is there any suggestions and plans?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,6 +76,7 @@ export default function loader(
   const {
     mode,
     target,
+    request,
     sourceMap,
     rootContext,
     resourcePath,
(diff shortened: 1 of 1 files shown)


dmitryuk:
I also need this PR, can it be merged?

PR #1809, opened 2021-03-17
Allow "layer" and "issuerLayer" rule properties

stevethedev:
The "layer" and "issuerLayer" rules are useful experimental properties that are currently stripped out by the Vue Loader plugin. This change fixes that issue.

My specific use-case was to load SVG files as Vue components if they are imported into a script, but import them as URLs if I reference them in the CSS. Layers make this trivial:

```
module.exports = {
  experiments: { layers: true },
  module: {
    rules: [
      { test: /\.vue$/, loader: "vue-loader" },
      { test: /\.vue$/, resourceQuery: /type=script/, layer: "script" },
      { test: /\.[jt]s$/, layer: "script" },
      {
        test: /\.svg$/i,
        oneOf: [
          { loader: 'vue-svg-loader', issuerLayer: "script" },
          { loader: 'file-loader' },
        ],
      },
    ],
  }
};
```

### DIFF
--- a/lib/plugin-webpack5.js
+++ b/lib/plugin-webpack5.js
@@ -20,12 +20,14 @@ const ruleSetCompiler = new RuleSetCompiler([
   new BasicMatcherRulePlugin('realResource'),
   new BasicMatcherRulePlugin('issuer'),
   new BasicMatcherRulePlugin('compiler'),
+  new BasicMatcherRulePlugin('issuerLayer'),
   new DescriptionDataMatcherRulePlugin(),
   new BasicEffectRulePlugin('type'),
   new BasicEffectRulePlugin('sideEffects'),
   new BasicEffectRulePlugin('parser'),
   new BasicEffectRulePlugin('resolve'),
   new BasicEffectRulePlugin('generator'),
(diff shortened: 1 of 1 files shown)


mrleblanc101:
Will this work in webpack 4 ? I'm trying to do something similar in Nuxt.
The problem is that `layer` and `issueLayer` do not exist in webpack 4, only resourceQuery.
```
config.module.rules.push({
    test: /\.svg$/,
    resourceQuery: /script/,
    use: ['babel-loader',
    'vue-svg-loader']
});
```
I couldnt' figure out why it's not working but when I look into the bundle code, there is no `?vue&type=template` so I think that's why my resourceQuery does nothing.
`__webpack_require__(/*! @/assets/svg/placeholder/organismes.svg*/ \"./src/assets/svg/placeholder/organismes.svg\")`
should be `__webpack_require__(/*! @/assets/svg/placeholder/organismes.svg?vue&type=template*/ \"./src/assets/svg/placeholder/organismes.svg?vue&type=template\")` I think for this to work.
I hope this fix will add these attributes back so that resourceQuery will work in Webpack 4

PR #1811, opened 2021-03-22
fix: Don't use rules that also match ".html"

### LINKED ISSUE #1625: vue-loader errored when html-loader is defined BEFORE it in webpack 5 configuration.
### Version
15.8.3

### Reproduction link
[https://github.com/ryanelian/vue-loader-bug](https://github.com/ryanelian/vue-loader-bug)

### Steps to reproduce
Add this line before vue-loader in webpack 5 configuration.

```
            {
                test: /\.html?$/,
                loader: 'html-loader'
            },
```

### What is expected?
No error.

⬢ webpack: Error: [VueLoaderPlugin Error] No matching use for vue-loader is found.
Make sure the rule matching .vue files include vue-loader in its use.

---
This is almost exactly the same as a past issue: https://github.com/vuejs/vue-loader/issues/1246
But this time it's webpack 5.

ubipo:
This adds a check so that rules that match foo.vue.html but also foo.html are ignored.

Resolves #1625

All tests pass and I've verified this myself with Webpack 5. I also added a Webpack 4 version but didn't verify that one.

I don't totally know if my assumption that a vue-loader rule should only match "foo.vue.html" and not "foo.vue" is correct...
Where is "foo.vue.html" used? I only ever use "foo.vue" myself.

### DIFF
--- a/lib/plugin-webpack4.js
+++ b/lib/plugin-webpack4.js
@@ -31,7 +31,12 @@ class VueLoaderPlugin {
     // find the rule that applies to vue files
     let vueRuleIndex = rawRules.findIndex(createMatcher(`foo.vue`))
     if (vueRuleIndex < 0) {
-      vueRuleIndex = rawRules.findIndex(createMatcher(`foo.vue.html`))
+      const vueHtmlRuleIndex = rawRules.findIndex(createMatcher(`foo.vue.html`))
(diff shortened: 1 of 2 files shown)


ubipo:
Huh weird, the tests pass on my machine

PR #1820, opened 2021-04-07
chore: adjust import statement

JayFate:
Signed-off-by: pengjie <37610029@qq.com>


src/descriptorCache.ts
```diff
@@ -1,6 +1,5 @@
 import * as fs from 'fs'
-import { SFCDescriptor } from '@vue/compiler-sfc'
-import { parse } from '@vue/compiler-sfc'
+import { SFCDescriptor, parse } from '@vue/compiler-sfc'
```
sodatea, line 2: It's better to change the first statement to `import type` than to merge them. They are different types of imports.

sodatea (changes requested):

PR #1824, opened 2021-04-23
fix(pitcher): css-loader >= v4 compat

Airkro:
fix css module with css-loader `options.esmodule: true`

fix: https://github.com/vuejs/vue-loader/issues/1788, https://github.com/vuejs/vue-loader/issues/1794, https://github.com/vuejs/vue-loader/issues/1709

PR #1867, opened 2021-08-22
fix: css module injectStyles bug in ssr close #1866

### LINKED ISSUE #1866: css  modules InjectStyles bug in ssr
### Version
16.4.1

### Reproduction link
[https://github.com/zhangyuang/vue-loader-ssr-bug.git](https://github.com/zhangyuang/vue-loader-ssr-bug.git)

### Steps to reproduce
yarn && npm start

### What is expected?
`this.$style` can be read the exact value in server and client side。

In server side `this.$style` value is `{default : val}`
In client side `this.$style` value is exact value

---
By read the source code, vue-loader must be used with vue-style-loader to avoid the problem.But i needn't use vue-style-loader。Maybe vue-loader should fix the case by itself

zhangyuang:
relate #1866

### DIFF
--- a/lib/codegen/styleInjection.js
+++ b/lib/codegen/styleInjection.js
@@ -91,6 +91,7 @@ module.exports = function genStyleInjectionCode (
         const request = genStyleRequest(style, i)
         styleInjectionCode += (
           `var style${i} = require(${request})\n` +
+          `style${i} = style${i}.default || style${i}\n` +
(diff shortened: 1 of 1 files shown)


zhangyuang:
The ci failed maybe not caused by the pr change

Airkro:
Install `memfs@3.2.3` may help.

https://github.com/vuejs/vue-loader/pull/1824#issuecomment-835769379

PR #1896, opened 2021-11-03
Continue looking for vue rule if not the first rule encountered

rfox12:
I believe "continue" was intended here instead of "break".  Unless there is a good reason the vue rule should be the first non-enforce rule.

https://githubmemory.com/repo/vuejs/vue-loader/issues/1826

### DIFF
--- a/lib/plugin-webpack5.js
+++ b/lib/plugin-webpack5.js
@@ -85,7 +85,7 @@ class VueLoaderPlugin {
           )
         }
         rawVueRules = rawRule
-        break
+        continue
       }
     }
     if (!vueRules.length) {


jherdman:
I don't believe this is correct either. I'm working on a project that has some Vue code, and some plain JS. If I implement your change I see this despite having some Vue files in play:

```
[webpack-cli] Error: [VueLoaderPlugin Error] No matching rule for .vue files found.
Make sure there is at least one root-level rule that matches .vue or .vue.html files
```

Your change, however, does get me further down the road in solving the error in the noted lined tickets.

jherdman:
I've tracked this down to a problem in my own configuration. I have a rule supplied by Shakapacker that looks like this:

```javascript

  test: /\.html$/,
  exclude: /\.(js|mjs|jsx|ts|tsx)$/,
  type: 'asset/source'
}
```

This rule trips the `foo.vue.html` test, which, in turn, confuses vue-loader.

I'm not a Vue expert by any stretch of the imagination, so I'm unclear if the `foo.vue.html` test is valid or not, but it seems like I need to intercept the rule above and improve the exclude regex.

Good luck on your own journey!

PR #1899, opened 2021-11-17
refactor: use refTransform instead of deprecated refSugar

cexbrayat:
Since https://github.com/vuejs/vue-next/commit/0805abe573de1ac726a0600077951ff3885653e6 the `refSugar` option has been deprecated,
and replaced with a `refTransform` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ## v16 Only Options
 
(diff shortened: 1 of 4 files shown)


cexbrayat:
No longer necessary with vue-loader v17

PR #1901, opened 2021-11-20
docs: recommend v-deep first

### LINKED ISSUE #1652: Remove /deep/ from documentation for future releases for Deep Selectors
Chrome will no longer throw a warning regarding using /deep/ if the documentation does not suggest using it as a solution. ::v-deep should be the de facto solution.

https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors

https://developers.google.com/web/updates/2017/10/remove-shadow-piercing

Some pre-processors, such as Sass, may not be able to parse >>> properly. In those cases you can use the ::v-deep combinator instead - it is an alias for >>> and works exactly the same.

KaelWD:
closes #1652

### DIFF
--- a/docs/guide/scoped-css.md
+++ b/docs/guide/scoped-css.md
@@ -48,11 +48,11 @@ With `scoped`, the parent component's styles will not leak into child components
 
 ## Deep Selectors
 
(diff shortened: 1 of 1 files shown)


mesqueeb:
@KaelWD
according to the Vue 3 docs, the preferred way is now `:deep` instead of `::v-deep`. Should we update the PR?

PR #1928, opened 2022-02-10
fixed comments

haochenli:


haochenli:
the comments should be 'except' rather than 'expect'

sodatea:
Thanks for the PR!
I've already merged #2010 before I saw this one, so I have to close it now. Thanks anyway!

PR #2006, opened 2022-10-13
Fix peer warning for webpack

ankurk91:
`pnpm` and `yarn` does not consider webpack version valid

Screenshot from 2022-10-13 11-06-56


ankurk91:
Looks like vue team does not care about PRs

PR #2046, opened 2023-05-20
feat: support experimental inline match resource

h-a-n-a:
This PR leverages inline match resource, a new technique provided by Webpack@5 
to process modules, which allows loader developers to write more ergonomic code
and make the code easier to maintain. From now on, you can enable 
`options.experimentalInlineMatchResource` of `vue-loader` to give it a try.

Other minor changes: 
1. Dispatch `VueLoaderPlugin` dynamically to decouple webpack and support webpack-like plugin systems.
2. Webpack@5 `experiments.css` support

Note:
The ideal solution with inline match resource is to get rid of the `VueLoaderPlugin` completely, 
I will try to optimize this in the future.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,7 +20,12 @@ import { formatError } from './formatError'
 import VueLoaderPlugin from './plugin'
 import { canInlineTemplate } from './resolveScript'
 import { setDescriptor } from './descriptorCache'
-import { getOptions, stringifyRequest as _stringifyRequest } from './util'
+import {
+  getOptions,
+  stringifyRequest as _stringifyRequest,
+  genMatchResource,
+  testWebpack5,
+} from './util'
 
 export { VueLoaderPlugin }
 
@@ -51,6 +56,7 @@ export interface VueLoaderOptions {
   exposeFilename?: boolean
   appendExtension?: boolean
   enableTsInTemplate?: boolean
+  experimentalInlineMatchResource?: boolean
 
   isServerBuild?: boolean
 }
@@ -90,18 +96,23 @@ export default function loader(
     rootContext,
     resourcePath,
     resourceQuery: _resourceQuery = '',
+    _compiler,
   } = loaderContext
 
+  const isWebpack5 = testWebpack5(_compiler)
   const rawQuery = _resourceQuery.slice(1)
   const incomingQuery = qs.parse(rawQuery)
   const resourceQuery = rawQuery ? `&${rawQuery}` : ''
   const options = (getOptions(loaderContext) || {}) as VueLoaderOptions
+  const enableInlineMatchResource =
+    isWebpack5 && Boolean(options.experimentalInlineMatchResource)
 
   const isServer = options.isServerBuild ?? target === 'node'
   const isProduction =
     mode === 'production' || process.env.NODE_ENV === 'production'
 
   const filename = resourcePath.replace(/\?.*$/, '')
+
   const { descriptor, errors } = parse(source, {
     filename,
     sourceMap,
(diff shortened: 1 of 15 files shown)


h-a-n-a:
@sodatea Hi, I haven't finished the final touch yet. But I have some questions about how this new feature will be integrated into the codebase. 
This feature, to me, is supposed to be enabled via some options like `experimentalInlineMatchResource` or something like this. However, this function affects both the implementation of plugin and loader, so I guess it will be better to configure in both plugin and loader(from the perspective of developer), or there might be some hacks. Which one do you prefer?
Another question is the option `experiments.css` provided by Webpack. Currently, I decide to detect the option automatically, is it necessary to provide another option for that?
Thank you for the help in advance!

sodatea:
> Which one do you prefer?

I'm okay with configuring it in both the plugin and loader.
Most users won't need to configure the loader/plugin by themselves anyway.

> Currently, I decide to detect the option automatically, is it necessary to provide another option for that?

I think automatic detection is better.

h-a-n-a:
@sodatea 
Firstly, thank you for the help! This PR is ready to review. I decide to leverage the old hack to make all possible configuration to `vue-loader` itself. So `options.experimentalInlineMatchResource` will only be exposed to the loader option (I think that will be enough as most of the user only want compilation-wise options). 
Secondly, the old rule modification will work with either inline match resource enabled or not, but only the new rule modification will apply with inline match resource enabled.
One more thing, webpack's `experiments.css` has some limitations with the current implementation, as `inline` and `module` will not work (requires `javascript`), however `scoped` is working as intended. Hope I can add some tests in the future PR.

src/index.ts
```diff
@@ -76,18 +77,26 @@ export default function loader(
     rootContext,
     resourcePath,
     resourceQuery: _resourceQuery = '',
+    _compiler,
   } = loaderContext
 
+  const webpackVersion = _compiler?.webpack?.version
+  const isWebpack5 = webpackVersion && webpackVersion[0] > '4'
```
hardfist, line 84: webpackVersion[0] > '4' is dangerous when webpackVersion > 10.x.y, I thinks direct equal test should be better

h-a-n-a: This is copied from the original implementation. I changed it to a relatively better one.

h-a-n-a:
@sodatea Hi, CI workflow has been fixed. Would you please take a look again? :)

src/util.ts
```diff
@@ -163,3 +164,38 @@ export function stringifyRequest(
       .join('!')
   )
 }
+
+export function genMatchResource(
+  context: LoaderContext<VueLoaderOptions>,
+  resourcePath: string,
+  resourceQuery?: string,
+  lang?: string,
+  additionalLoaders?: string[]
```
sodatea, line 173: I don't see any usage of this parameter. What's it reserved for?

h-a-n-a: This is not related. I will remove this.

src/pluginWebpack5.ts
```diff
@@ -221,16 +223,53 @@ class VueLoaderPlugin {
...
+          loader === require('../package.json').name ||
+          loader.startsWith(require.resolve('./index'))
+        )
+      })
+
+      compiler.options.module!.rules = [
+        pitcher,
+        ...rules.filter((rule) => !vueLoaderRules.includes(rule)),
+        templateCompilerRule,
+        ...clonedRules,
```
sodatea, line 261: I thought the intention of using inline match resource was to get rid of cloning rules. Can we remove it here?

h-a-n-a: I think what blocks us from removing this currently is the `lang` query for us. `templateCompilerRule` seems depending on the result from the remaining rules. But I will try this in the future PR. Ideally, we should get rid of the `VueLoaderPlugin` in the future.

sodatea (approved):

PR #2058, opened 2023-09-28
feat: v15 support experimental inline match resource

jeromehan:
I think vue-loader@15.x supports webpack5 experiments.css just as important, there are many people who use vue2.x, you might consider this pr @sodatea


sodatea (changes requested):

sodatea:
Thanks for the help! Now published as https://github.com/vuejs/vue-loader/releases/tag/v15.11.0

PR #2060, opened 2023-10-18
fix: fix read property error for accessing lang of script

### LINKED ISSUE #2059: Build error: Cannot read properties of null (reading 'lang')
After update to v15.11.0, i'm facing error: Cannot read properties of null (reading 'lang'). Everything is fine, if i downgrade to v[15.10.2](https://www.npmjs.com/package/vue-loader/v/15.10.2)

freedomlang:
fix #2059

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -117,7 +117,7 @@ module.exports = function (source) {
   // let isTS = false
   const { script, scriptSetup } = descriptor
   if (script || scriptSetup) {
-    const lang = script?.lang || scriptSetup?.lang
+    const lang = (script && script.lang) || (scriptSetup && scriptSetup.lang)
     // isTS = !!(lang && /tsx?/.test(lang))
     const externalQuery =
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -117,7 +117,7 @@ module.exports = function (source) {
   // let isTS = false
   const { script, scriptSetup } = descriptor
   if (script || scriptSetup) {
-    const lang = script.lang || (scriptSetup && scriptSetup.lang)
+    const lang = script?.lang || scriptSetup?.lang
```
sodatea, line 120:
```suggestion
    const lang = (script && script.lang) || (scriptSetup && scriptSetup.lang)
```

Thanks for the quick fix!
However, v15's source code is untranspiled JavaScript and may run in environments that do not support optional chaining syntax. So here we must use the logical OR operator instead.

sodatea (approved):

PR #2079, opened 2024-07-19
fix: missing register component in ssr (#1887)

mog422:

PR #2083, opened 2024-08-07
fix: add tests for SSR render

niksy:
This PR continues on the work of https://github.com/vuejs/vue-loader/pull/2079 by adding missing tests removed in https://github.com/vuejs/vue-loader/commit/20352da36bf5e37b761ab48971f25f194e6dac7c and [based on implementation at v15 version of loader](https://github.com/vuejs/vue-loader/blob/master/test/ssr.spec.js).

SSR context styles are not present since I couldn’t get final result inside context. As I see it, that currently isn’t possible since all the CSS is automatically extracted, but maybe it could be achivede with additional loader after all the CSS has been preprocessed?

v15 version relies on vue-style-loader so I think that’s why it was possible to get all the necessary information.

Also, hashing function from original PR has been moved from `hash-sum` to internal one to align with Vite and other hashing behavior. It’s possible that this could clash with [Nuxt Vue client manifest package implementation](https://github.com/nuxt/nuxt/blob/main/packages/webpack/src/plugins/vue/client.ts).

@sodatea Is it possible we can get this implemented? This would greatly help with collecting all the necessary components needed for initial render in SSR contexts. After that, maybe we could work on critical CSS implementation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   "packageManager": "pnpm@8.12.0",
   "dependencies": {
     "chalk": "^4.1.0",
+    "hash-sum": "^2.0.0",
     "watchpack": "^2.4.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


OskarLebuda:
Good job 👍🏻  I've done it myself but this requires fork of this repo. 
The official way should be provided.  @yyx990803  @haoqunjiang  approve this PR and make some release with that?

dmitryuk:
Thanks for you PR! Waiting to be merged soon

PR #2092, opened 2025-07-24
feat: support multiple css-modules imports

escaton:
Currently `vue-loader` doesn't support multiple `css module` sections.
In this example the last one will rewrite the `$style` object, and `$style.root` will be `undefined`.
```vue
<template>
  <div :class="$style.root">
    <div :class="$style.child />
  </div>
</template>

<style module critical>
.root { background: black; }
</style>

<style module>
.child { background: red; }
</style>
```
This PR fixes that.

### DIFF
--- a/src/cssModules.ts
+++ b/src/cssModules.ts
@@ -1,3 +1,13 @@
+export function genInitCSSModulesCode() {
+  return `
+const cssModules = {}
+const cssModulesStore = {}
+const getCssModules = (name) => {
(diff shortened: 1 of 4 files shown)


escaton:
Found a duplicate PR: https://github.com/vuejs/vue-loader/pull/2045 with slightly different approach to HMR, but a lot more tests

escaton:
@yyx990803 @Jinjiang calling you directly, as the repo seems almost abandoned. Could you please take a look at this or #2045 and merge one of them?

